PR #9, opened 2026-03-18
[JBRes-8664] Fix handling of database migration locks

samorojy:
This PR adds proper handling of migration locks. If migrations are started by one worker, other workers will wait until they are finished.

Resolves: JBRes-8664

### DIFF
--- a/orchestrator/src/idegym/orchestrator/database/database.py
+++ b/orchestrator/src/idegym/orchestrator/database/database.py
@@ -1,3 +1,4 @@
+import asyncio
 from contextlib import asynccontextmanager
 from typing import Any, List, NamedTuple, Optional, Set, cast
 from uuid import UUID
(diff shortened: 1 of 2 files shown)


vpoliakov-pixel (approved):

orchestrator/src/idegym/orchestrator/database/database.py
```diff
@@ -61,7 +62,43 @@ async def init_db(db_url: str, config: SQLAlchemyConfig, clean_database: bool =
...
+    if not ran_migrations:
+        logger.info("Waiting for migrations to complete...")
+        max_wait_time = 300  # 5 minutes timeout
+        poll_interval = 1  # Check every second
+        elapsed = 0
+
+        # Get the expected migration version from Alembic
+        expected_version = migration_manager.get_expected_version()
+
+        while elapsed < max_wait_time:
```
dabico: Are we limiting by max execution time or iterations? Because `elapsed` is only increased on each new iteration. As a result, when this runs for 300 iterations max, it may be more time spent in reality. If you want to accurately limit by time, maybe it's better to use `asyncio.timeout` context manager with an infinite loop inside?

samorojy: That's true. It was a part of the suggested patch file linked to the ticket. 

Fixed! 

Used asyncio loop time for better time tracking.

PR #10, opened 2026-03-19
[JBRes-4975] update `uv` to the latest version

vpoliakov-pixel:
Resolves: JBRes-4975


dabico:
I think we need to update the CI workflows as well. The latest version of the `uv` action came with the feature to detect the version from `pyproject.toml`. We can upgrade that step and then drop the specified version in the workflow. I also think we need to update which version of `uv` use in Docker image builds, because right now it's still the old one.

vpoliakov-pixel:
s u c c e s s

Dockerfile.jinja
```diff
@@ -65,16 +65,16 @@ USER appuser
...
+COPY --chown=appuser:appuser api api/
+COPY --chown=appuser:appuser backend-utils backend-utils/
+COPY --chown=appuser:appuser common-utils common-utils/
+COPY --chown=appuser:appuser rewards rewards/
+COPY --chown=appuser:appuser tools tools/
+COPY --chown=appuser:appuser server server/
 
 RUN set -eux; \
-    uv python install; \
+    uv python install 3.12; \
```
dabico: I just realized we don't `COPY` the `.python-version` file from our repository. This way, we don't need to hard-code the versioning requirement in the `Dockerfile`. Let's copy it and just revert to how it was before?

dabico (commented):
It's a bit weird that we use two different build systems now for the orchestrator and server. Also, please remember to update the version of `uv` in `pre-commit` hooks configuration.

dabico (approved):
This looks great now, thanks for resolving this!

.github/workflows/ci.yml
dabico: Actually, can you also update the `ruff` version for actions? We can also fix it in a subsequent PR.

PR #12, opened 2026-03-19
[JBRes-6036] clean up orphaned kaniko jobs; better error handling during status monitoring

vpoliakov-pixel:
Resolves: JBRes-6036

### DIFF
--- a/orchestrator/src/idegym/orchestrator/watcher.py
+++ b/orchestrator/src/idegym/orchestrator/watcher.py
@@ -171,11 +171,7 @@ async def check_orphaned_kaniko_jobs(db: AsyncSession, namespace: str):
                     f"Orphaned job detected: '{job_name}' is IN_PROGRESS in DB but {k8s_status} in k8s. Updating..."
                 )
                 await update_job_status(
-                    db,
-                    job_name,
-                    status=k8s_status,
-                    tag=job_record.tag,
-                    request_id=job_record.request_id
+                    db, job_name, status=k8s_status, tag=job_record.tag, request_id=job_record.request_id
                 )
                 logger.info(f"Updated orphaned job '{job_name}' status to {k8s_status}")
         except Exception:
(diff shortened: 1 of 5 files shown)


orchestrator/src/idegym/orchestrator/watcher.py
```diff
@@ -208,13 +247,15 @@ async def cleanup_inactive_pods(watcher_config: WatcherConfig):
 
             try:
                 logger.info("Starting cleanup operations with advisory lock acquired")
+                namespace = env.get("__NAMESPACE", "idegym")
```
dabico: When I originally created these environment variables, I used the `__` prefix to signal that they were to be used within the scope of the deployment template (for the OTel config). But since we've started using them, I think we should change them to the `IDEGYM_` prefix like we use for everything else. We can do this at some later point though.

orchestrator/src/idegym/orchestrator/watcher.py
dabico, line 1: While the `Job` resource has a TTL, the `ConfigMap` associated with it won't get cleaned up, regardless of the outcome. We could start using `ownerReferences` to link `ConfigMap` to `Job`. Then, it would get properly cleaned up. What do you think?

dabico (approved):

orchestrator/src/idegym/orchestrator/watcher.py
dabico: Ah, I see the issue now: JBRes-8762.

vpoliakov-pixel:
Renamed env variable.
Fixed cleaning of pdbs and config maps for kaniko jobs.

Lets implement ownerReferences approach for the whole thing, both kaniko jobs and IDEGym server's deployment-service-pod-pdb separately

orchestrator/src/idegym/orchestrator/watcher.py
dabico: Created: JBRes-8772

PR #32, opened 2026-03-23
[JBRes-8318] return finished server in fifo order for new requests

vpoliakov-pixel:
Resolves: JBRes-8318


orchestrator/src/idegym/orchestrator/database/helpers.py
```diff
@@ -145,22 +145,36 @@ async def check_resources_and_save_server_in_db(
...
-    existing_server = await find_matching_finished_server(
-        db, client_name, request.server_name, request.image_tag, request.runtime_class_name, request.run_as_root
+    lookup_result = await find_matching_finished_server(
+        db,
+        client_name,
+        request.server_name,
+        request.image_tag,
+        request.runtime_class_name,
+        request.run_as_root,
+        enable_fifo_check=enable_fifo_check,
```
dabico, line 165: I'd suggest using explicit argument names here to be consistent with other changes:

```suggestion
    lookup_result = await find_matching_finished_server(
        db=db,
        client_name=client_name,
        server_name=request.server_name,
        image_tag=request.image_tag,
        container_runtime=request.runtime_class_name,
        run_as_root=request.run_as_root,
        enable_fifo_check=enable_fifo_check,
```

orchestrator/src/idegym/orchestrator/database/database.py
```diff
@@ -323,40 +324,128 @@ async def get_idegym_servers_by_status(db: AsyncSession, statuses: Set[Availabil
...
 
 
+async def has_pending_start_server_operations(
+    db: AsyncSession,
+    client_name: str,
+    image_tag: str,
+    container_runtime: str,
+    run_as_root: bool,
+    server_name: Optional[str] = None,
+    scheduled_before: int = None,
```
dabico: Wrong type hint:

```suggestion
    scheduled_before: Optional[int] = None,
```

orchestrator/hydra_configs/orchestrator/base.yaml
```diff
@@ -41,3 +41,5 @@ watcher:
   finished_timeout: ${oc.env:IDEGYM_WATCHER_FINISHED_TIMEOUT, "PT5M"}
   request_max_age: ${oc.env:IDEGYM_WATCHER_REQUEST_MAX_AGE, "P14D"}
   request_stale: ${oc.env:IDEGYM_WATCHER_REQUEST_STALE, "PT24H"}
+
+enable_fifo_server_reuse: ${oc.env:IDEGYM_ENABLE_FIFO_SERVER_REUSE, "False"}
```
dabico, line 45: Maybe I missed this, but what was the idea behind putting this behind a feature flag? Are there cases where FIFO is not desirable?

dabico (approved):

orchestrator/hydra_configs/orchestrator/base.yaml
vpoliakov-pixel: It adds at least one more database query per server creation request and it could slow the things down. It is easy to wrap it under the feature flag and test performance / fix performance separately

PR #33, opened 2026-03-25
[JBRes-8782] Add full integration tests in minikube

vpoliakov-pixel:
Resolves: JBRes-8782


pyproject.toml
```diff
@@ -56,7 +57,7 @@ members = [
 [tool.pytest.ini_options]
 python_files = ["test_*.py"]
 python_functions = ["test_*"]
-pythonpath = ["."]
+pythonpath = [".", "local-testing"]
```
vpoliakov-pixel: probably not needed anymore

orchestrator/src/idegym/orchestrator/router/server.py
```diff
@@ -174,6 +175,10 @@ async def _task_start_server(config: Config, request: StartServerRequest, async_
 
             # For RESTART, we'll mark ALIVE later below; for RESET we skip marking ALIVE
         else:
+            if client_name is None:
```
vpoliakov-pixel, line 182: need to double check this fix

local-testing/README.md
```diff
@@ -0,0 +1,279 @@
+# Local Testing with Minikube
```
vpoliakov-pixel: need to update readme

uv.lock
```diff
@@ -17,6 +17,7 @@ members = [
     "idegym-rewards",
     "idegym-server",
     "idegym-tools",
+    "local-testing",
```
dabico: Let's have the workspace name follow the convention we laid out so far and use the `idegym-` prefix.

local-testing/utils/k8s_setup.py
```diff
@@ -0,0 +1,363 @@
...
+from idegym.utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+BASE_URL = "http://idegym-local.test"
+
+
+def get_config_dir() -> Path:
+    """Get the config directory containing kustomization.yaml."""
+    return Path(__file__).parent.parent / "config"
```
dabico: If you add `__init__.py` to `config`, then you can treat it as a resource directory and import it using `importlib.resources`. IMO this beats the current approach, since it will survive refactoring. See: https://github.com/JetBrains-Research/idegym/blob/6cbd071cb280037d7fe155c2390158a184668522/backend-utils/src/idegym/backend/utils/bash_executor.py#L7-L17

e2e-tests-minikube/utils/k8s_setup.py
dabico, line 1: I see a lot of manual `kubectl` calls, but we import `kubernetes-asyncio` as a dependency. Why not make use of it then?

local-testing/tests/test_server_strategies.py
```diff
@@ -0,0 +1,255 @@
...
+            # File operations
+            await server.create_file(file_path="/tmp/test.py", content="print('hello')")
+            result = await server.execute_bash(script="python /tmp/test.py")
+            assert "hello" in result.stdout
+
+
+@pytest.mark.asyncio
+async def test_reward_operations(test_image):
+    """Test setup, compilation, and test rewards."""
+    test_id = uuid4().hex[:8]
```
dabico: By generating a UUID, then discarding most of the information, you essentially discard a part of the generated randomness. If you want a truly unique alphanumeric sequence, then it might be better to use:

```suggestion
    test_id = urandom(4).hex()
```

Note that you need to:

```python
from os import urandom
```

local-testing/tests/test_server_strategies.py
```diff
@@ -0,0 +1,255 @@
...
+                close_action=ServerCloseAction.STOP,
+            ) as s2,
+        ):
+            assert s1.server_id != s2.server_id
+
+
+@pytest.mark.asyncio
+async def test_bash_and_file_operations(test_image):
+    """Test bash execution and file operations."""
+    test_id = uuid4().hex[:8]
```
dabico: Considering these are defined in multiple functions, I think we can extract them into [fixtures](https://docs.pytest.org/en/stable/explanation/fixtures.html).

local-testing/tests/test_server_strategies.py
```diff
@@ -0,0 +1,255 @@
+"""Test different server reuse strategies and advanced operations."""
+
+from uuid import uuid4
+
+import pytest
+from idegym.api.orchestrator.servers import ServerReuseStrategy
+from idegym.client.client import ServerCloseAction
+from kubernetes_asyncio.client import V1ResourceRequirements
+
+from .idegym_utils import create_http_client
```
dabico: Not sure about this import. We don't use notation like this in the rest of the project.

local-testing/tests/test_annotated_types.py
dabico, line 1: These were migrated to unit test to check the validator functionality:

https://github.com/JetBrains-Research/idegym/blob/6cbd071cb280037d7fe155c2390158a184668522/unit-tests/test_types.py#L1-L68

I think running these in an e2e/integration test suite is a bit overkill.

local-testing/tests/conftest.py
```diff
@@ -0,0 +1,381 @@
...
+    commands = """
+USER root
+RUN set -eux; \\
+    apt-get update; \\
+    apt-get install -y --no-install-recommends \\
+    python3=3.11.2* \\
+    python-is-python3=3.11.2*; \\
+    rm -rf /var/lib/apt/lists/*
+USER appuser
+"""
```
dabico: Why not have this as a `Dockerfile` in `config`? We already load some resources from that directory in this file.

local-testing/tests/conftest.py
```diff
@@ -0,0 +1,381 @@
...
+
+import subprocess
+from pathlib import Path
+
+import pytest
+from idegym.api.docker import BaseImage
+from idegym.api.git import GitRepositorySnapshot
+from idegym.client import IdeGYMDockerAPI
+from idegym.utils.logging import get_logger
+from utqils.k8s_setup import wait_for_service
```
dabico: Think there's a typo here that would cause a fail:


```suggestion
from utils.k8s_setup import wait_for_service
```

dabico (changes requested):
Will do another pass later but here are some changes to start with.

e2e-tests-minikube/pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[project]
+name = "idegym-e2e-tests"
+version = "0.1.0"
```
dabico: We used:

```suggestion
version = "0.0.0.dev0"
```

For all workspaces.

e2e-tests-minikube/utils/build_images.py
```diff
@@ -0,0 +1,147 @@
...
+
+from idegym.utils.logging import get_logger
+from jinja2 import Template
+
+logger = get_logger(__name__)
+
+
+def get_repo_root() -> Path:
+    """Get the repository root directory."""
+    return Path(__file__).parent.parent.parent
```
dabico: We have the `from_root` package in our development dependencies. Used it in the past. Example: https://github.com/JetBrains-Research/idegym/blob/7d60a6308df4185a89f8461e07bd02851206b39c/scripts/build_orchestrator_image.py#L88-L94

e2e-tests-minikube/utils/build_images.py
```diff
@@ -0,0 +1,147 @@
...
+    logger.info("Building orchestrator image...")
+
+    repo_root = get_repo_root()
+    script_path = repo_root / "scripts" / "build_orchestrator_image.py"
+
+    if not script_path.exists():
+        raise FileNotFoundError(f"Build script not found: {script_path}")
+
+    # Use uv run to execute the script with its declared dependencies
+    cmd = ["uv", "run", str(script_path), "--versions", "latest"]
```
dabico: The script is executable and has a shebang:

https://github.com/JetBrains-Research/idegym/blob/7d60a6308df4185a89f8461e07bd02851206b39c/scripts/build_orchestrator_image.py#L1-L3

You can run it directly.

e2e-tests-minikube/utils/build_images.py
```diff
@@ -0,0 +1,147 @@
...
+
+    logger.info(f"Running: {' '.join(cmd)}")
+    result = subprocess.run(cmd, cwd=repo_root, check=True)
+
+    if result.returncode == 0:
+        logger.info("✓ Orchestrator image built successfully")
+    else:
+        raise RuntimeError(f"Failed to build orchestrator image: exit code {result.returncode}")
```
dabico: Why do you need to do this if `check=True`? It will throw a `CalledProcessError` if exit code is non-zero.

e2e-tests-minikube/utils/build_images.py
```diff
@@ -0,0 +1,147 @@
...
+    dockerfile_template_path = repo_root / "Dockerfile.jinja"
+    with open(dockerfile_template_path, "r") as f:
+        template = Template(f.read())
+
+    rendered_dockerfile = template.render(repository="docker.io/library", image="debian", tag="bookworm-20250520-slim")
+
+    # Write rendered Dockerfile to a temporary file
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".Dockerfile", delete=False) as tmp:
+        tmp.write(rendered_dockerfile)
+        tmp_dockerfile_path = tmp.name
```
dabico: Why do `delete=False` and the cleanup separately? You can remove the `try`-`finally` completely and execute the inner context code as part of the `with`. If there is an exception, then the file will get deleted anyway on context manager exit.

e2e-tests-minikube/utils/build_images.py
dabico, line 1: BTW we still use `python_on_whales` as a dependency. Maybe we can make use of it instead of doing manual `subprocess` calls for `docker`. Recall that it wraps around the CLI.

e2e-tests-minikube/tests/test_server_strategies.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+@pytest.mark.asyncio
+async def test_resource_limits_enforcement(test_image, test_id):
+    """
+    Test that resource limits are enforced.
+    Try to create a server with excessive resources (100 CPU, 1000Gi memory).
+    Should fail with 429 or timeout after retrying 429 errors.
+    """
+    async with create_http_client(name=f"limits-{test_id}", nodes_count=0, request_timeout_in_seconds=60) as client:
```
dabico, line 135: We use [`pytest.raises`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.raises) in other places to assert exceptions.

e2e-tests-minikube/README.md
```diff
@@ -0,0 +1,358 @@
...
+## Prerequisites
+
+### 1. Install Required Tools
+
+Install Docker, kubectl, minikube, and uv:
+
+```bash
+brew install --cask docker
+brew install kubernetes-cli minikube
+brew install uv
```
dabico: Can be one step:


```suggestion
brew install kubernetes-cli minikube uv
```

e2e-tests-minikube/utils/idegym_utils.py
```diff
@@ -0,0 +1,74 @@
...
+        heartbeat_interval_in_seconds: Interval in seconds for sending heartbeats (default: 60)
+        request_timeout_in_seconds: Default timeout in seconds for every operation related to a http client (default: 60)
+        client_id: If provided, the client will work as a client with the given ID, but will not send heartbeats
+        **kwargs: Additional keyword arguments to pass to IdeGYMClient constructor
+    """
+    base_url, username, password = get_test_params()
+
+    # Remove trailing slash if present
+    if base_url.endswith("/"):
+        base_url = base_url[:-1]
```
dabico: Why not use the dedicated string manipulation function?


```suggestion
    base_url = base_url.rstrip("/")
```

dabico (changes requested):
More feedback. Will make another pass later. Please note that some comments apply to several places.

e2e-tests-minikube/config/kustomization.yaml
dabico, line 1: Do you think we should deploy all the monitoring services for tests? I don't think that's necessary, and we can just deploy the database and orchestrator. We aren't exactly testing metric or trace collection.

orchestrator/supervisord.conf
```diff
@@ -19,7 +19,7 @@ supervisor.rpcinterface_factory = supervisor.rpcinterface:make_main_rpcinterface
 process_name=orchestrator
 numprocs=1
 priority=2
-command=python -m idegym.orchestrator.main
+command=uv run python -m idegym.orchestrator.main
```
dabico, line 22: I'm not sure if we should rely on `uv` as a runtime dependency. Given that we `uv sync` as part of the build process, Python will be symlinked to `.venv` and we can call it directly:

```suggestion
command=.venv/bin/python -m idegym.orchestrator.main
```

However, I'm not sure if this requires changing the `PYTHONPATH` environment variable. Though considering we don't use something like a [multi-stage build](https://github.com/astral-sh/uv-docker-example/blob/00701a3361b9bdc896f8549fb9f890f99d115824/multistage.Dockerfile), I doubt that it makes a huge difference. At the end of the day, I guess it's what you prefer, I'm just talking out loud here.

e2e-tests-minikube/run_tests.py
```diff
@@ -0,0 +1,170 @@
...
+    Returns:
+        bool: True if tests passed, False otherwise
+    """
+    import subprocess
+
+    module_dir = Path(__file__).parent
+    tests_dir = module_dir / "tests"
+
+    # Override root pytest config to avoid dependency on pytest-randomly
+    cmd = [sys.executable, "-m", "pytest", str(tests_dir), "-v", "-s", "-o", "addopts="]
```
dabico: > Override root pytest config to avoid dependency on pytest-randomly

Why? Is it causing some kind of problem?

e2e-tests-minikube/run_tests.py
dabico, line 1: I hope that in the future we won't need a custom test runner and we will be able to launch all our tests from the repository root with just `pytest`. Though I think we need to set up some kind of CI with more compute to handle it. Default runners won't be sufficient, I believe.

e2e-tests-minikube/tests/conftest.py
```diff
@@ -0,0 +1,209 @@
...
+            pod_name = pod.metadata.name if pod.metadata and pod.metadata.name else ""
+            # Pod name format: {deployment-name}-{replicaset-hash}-{pod-hash}
+            parts = pod_name.rsplit("-", 2)
+            if len(parts) >= 3:
+                deployment_names.add(parts[0])
+
+    for deployment_name in deployment_names:
+        k8s_client.delete_deployment(namespace=DEFAULT_NAMESPACE, deployment_name=deployment_name)
+        k8s_client.delete_services(namespace=DEFAULT_NAMESPACE, service_names=[f"{deployment_name}-service"])
+        k8s_client.delete_pod_disruption_budget(namespace=DEFAULT_NAMESPACE, pdb_name=f"{deployment_name}-pdb")
```
dabico: With #35 merged, we'd no longer have to manually clean all these, just the `Deployment`:


```suggestion
    for deployment_name in deployment_names:
        k8s_client.delete_deployment(namespace=DEFAULT_NAMESPACE, deployment_name=deployment_name)
```

e2e-tests-minikube/tests/conftest.py
```diff
@@ -0,0 +1,209 @@
...
+        has_server_container = any(
+            container.name == SERVER_CONTAINER_NAME for container in containers if container.name
+        )
+
+        if has_server_container:
+            pod_name = pod.metadata.name if pod.metadata and pod.metadata.name else ""
+            # Pod name format: {deployment-name}-{replicaset-hash}-{pod-hash}
+            parts = pod_name.rsplit("-", 2)
+            if len(parts) >= 3:
+                deployment_names.add(parts[0])
```
dabico: When you create a `Deployment`, the `ownerReferences` are automatically set by the `ReplicaSet` controller: `Deployment` ← `ReplicaSet` ← `Pod`. Why not traverse those to find the `Deployment` name instead of this heuristic?

dabico: Alternatively, you can list all `Deployment` resources with `app.kubernetes.io/component=sandbox` label which I added in the same PR.

e2e-tests-minikube/README.md
dabico, line 1: This document has some duplicated information from `local_deployment.md`. Wonder if we can consolidate it somehow into one place. Or maybe mention specific sections from that document as prerequisites for reading this one.

e2e-tests-minikube/README.md
```diff
@@ -0,0 +1,356 @@
...
+Integration tests for IdeGYM that run on a local minikube cluster without requiring remote registry access.
+
+## Prerequisites
+
+### 1. Install Required Tools
+
+Install Docker, kubectl, minikube, and uv:
+
+```bash
+brew install --cask docker
```
dabico: By the way, this has changed since I originally wrote that other document. The cask is now called `docker-desktop`, and one should install either that or just the `docker` formula.

e2e-tests-minikube/utils/constants.py
```diff
@@ -0,0 +1,26 @@
+"""Shared constants for e2e testing."""
+
+# Kubernetes configuration
+DEFAULT_NAMESPACE = "idegym-local"
+INGRESS_NAMESPACE = "ingress-nginx"
+INGRESS_CONTROLLER_SERVICE = "ingress-nginx-controller"
+
+# URLs
+BASE_URL = "http://idegym-local.test"
```
dabico, line 9: Why use `idegym-local` instead of `idegym` like in the local deployment guide? Same for the URL.

dabico (changes requested):

e2e-tests-minikube/config/kustomization.yaml
samorojy: Could we use it for tests for logging? That all the expected fields are in place. 

(to prevent what we are still having  [8684] with missing fields in the logs)

dabico: We can check `Deployment` logs directly. We don't have any monitoring deployment for Loki or some other log aggregator (and we don't need it if I'm being honest).

> (to prevent what we are still having [8684] with missing fields in the logs)

But could you explain this? Is this about the missing timestamps?

samorojy: I think we might be missing a bit more than timestamps because there were some quick hacks :)

e2e-tests-minikube/utils/constants.py
vpoliakov-pixel: I like clarity and reusing names for different things is not my style.
`idegym` is for local runs where you check your features manually, use your own database, etc.
`idegym-local` is for auto tests only.
Same applies for the namespace too, in auto tests it could/should/would be cleared regularly but your own local setup could follow you for month.
I do agree that the naming is not the best though.

e2e-tests-minikube/README.md
vpoliakov-pixel: I have made it intentionally to avoid confusion of any kind.
I suggest restructuring docs in a separate PR.

e2e-tests-minikube/run_tests.py
vpoliakov-pixel: Today is the day, the future has come, now everything runs through pytest but e2e marker is temporary disabled for CI until we find the appropriate agent machines.

orchestrator/supervisord.conf
vpoliakov-pixel: could be changed later if needed, i do not see problems with following either way

e2e-tests-minikube/config/kustomization.yaml
vpoliakov-pixel: The idea is to make those tests as close to "real" env as possible.

It could be used for testing metrics, logs; if you want to write such tests.
It could be useful for debugging.
It could be useful to check if we don't break anything if play around configs for these services.

I do understand that keeping them requires more resources but it should not really be a problem in a modern world.

Or we can introduce a bunch of flags to deploy/not deploy them depending on smth. But that is another task.

dabico (approved):
Thank you!

PR #35, opened 2026-03-26
[JBRes-8762] Use owner references for easier resource cleanup

dabico:
### Summary

Improve how the orchestrator manages Kubernetes resources by introducing standard labels, owner references and autoscaler annotations across all dynamically created resources (Kaniko jobs, node holders and user sandboxes).

`ownerReferences` are now set on child resources (`ConfigMap`, `PodDisruptionBudget` & `Service`) pointing to their parent (`Deployment` or `Job`). This enables Kubernetes garbage collection to automatically clean up child resources when the parent is deleted, eliminating the need for manual cleanup logic. As a result, `clean_up_after_job` and the multi-resource deletion in `clean_up_server` have been removed. Further simplifications can be made, but I opted to leave them for another time.

### Labels

All dynamically created resources now carry standard `app.kubernetes.io/*` labels:

- `app.kubernetes.io/name`: resource name (mirrors `app`).
- `app.kubernetes.io/component`: role within the system (`image-builder`, `node-holder` and `sandbox`).
- `app.kubernetes.io/version`: version of IDEGym at creation time.
- `app.kubernetes.io/part-of: idegym`: groups all resources under the same application.

Label changes in static manifests:

- `app.kubernetes.io/component` was changed from `api` to `controller` for the `orchestrator`.
- `app.kubernetes.io/part-of` label was added to static YAML manifests (for `orchestrator` and `postgresql`).

### Annotations

Pod templates for sandboxes, Kaniko jobs and node holders now include:

```yaml
cluster-autoscaler.kubernetes.io/safe-to-evict: "false"
```

Its purpose is outlined [here](https://kubernetes.io/docs/reference/labels-annotations-taints/#cluster-autoscaler-kubernetes-io-safe-to-evict). In essence, it prevents the cluster autoscaler from evicting these workloads during scale-down. Its relation to the cluster autoscaler is further elaborated on [here](https://github.com/kubernetes/autoscaler/blob/3ac5fc596fe04b18dd764f35c766cda0cb4cd512/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node).

### Other changes

- **Simplify resource naming.** Forego suffixes like `-service` and `-pdb`. The resource kind already conveys what it is, and it's common to use matching names between related resources.
- **Pin Kaniko image to the final release:** `gcr.io/kaniko-project/executor:v1.24.0`.
- **Update Kaniko job naming.** Use `getrandbits` to generate a random name instead of `uuid4`.
- **Use a standardized pause image for node holders:** `registry.k8s.io/pause`. More efficient in terms of image size, resource usage and startup time.
- **Update node holder naming.** Use a fixed-length hash of the client name to avoid conflicts with resource name length restrictions imposed by Kubernetes.
- **Remove the `container_name` and `replicas_count` parameters from `deploy_server`.** They were unused since the method was introduced.
- **The `construct_forwarding_payload` was changed from `async` to a regular method.** There was no need for this as there was no blocking code in the implementation.
- **`Service` URL was updated to match the new naming scheme** (no `-service` suffix).
- **Removed debugging environment variables from orchestrator manifest**: `IDEGYM_ASYNCIO_DEBUG`, `IDEGYM_ASYNCIO_DUMP_INTERVAL` & `IDEGYM_LOG_LEVEL`.
- **Removed unused initializer from `ResourceDeletionFailedException`.** No longer needed as the logic is simplified.

Resolves: JBRes-8762


orchestrator/src/idegym/orchestrator/nodes_holder.py
```diff
@@ -59,93 +90,129 @@ async def spin_up_or_update_nodes_for_client(
...
-    deployment_name = f"{idegym_nodes_holder_prefix}-{client_name}"
-
-    # Set up resource requirements
-    resources = V1ResourceRequirements(
-        limits={"memory": "150Mi", "cpu": "100m"}, requests={"memory": "150Mi", "cpu": "100m"}
-    )
+    client_hash = md5(client_name.encode()).hexdigest()
+    name = f"{component}-{client_hash}"
+    annotations = {
+        "cluster-autoscaler.kubernetes.io/safe-to-evict": "false",
```
vpoliakov-pixel, line 95: do we still need pdbs with this approach?

vpoliakov-pixel (approved):

orchestrator/src/idegym/orchestrator/nodes_holder.py
```diff
@@ -162,36 +229,23 @@ async def release_nodes_for_client(
     """
     Delete the deployment and PodDisruptionBudget that were created for a client to hold its nodes.
     """
-    deployment_name = f"{idegym_nodes_holder_prefix}-{client_name}"
-    pdb_name = f"{deployment_name}-pdb"
-
+    client_hash = md5(client_name.encode()).hexdigest()
```
vpoliakov-pixel: can be extracted as helper function of some sort

dabico: Created utility in `common-utils`, and applied it to several places in the codebase.

orchestrator/src/idegym/orchestrator/nodes_holder.py
dabico: > do we still need pdbs with this approach?

It's hard to test this and get a definitive confirmation. This is because it's not possible to directly manipulate the autoscaler into scaling. A replicable test is not exactly easy to pull off, but I'm looking into some possibilities.

However, having researched the topic more, I'd like to highlight a distinction between the two in how they operate and what they do. Let's start from what we have: **pod disruption budgets prevent eviction events at the API level**. This means that even by using something like `kubectl drain` you won't be able to evict the workload off the node. On the other hand, the annotation only instructs the **cluster autoscaler not to evict the workload during scaling events**. However, eviction events are not blocked altogether, meaning that the workload will be scheduled elsewhere if you use `kubectl drain`. So the difference is that one will completely block evictions, while the other will only block evictions triggered by the cluster component responsible for scaling.

Having said that, you might be tempted to think that PDBs are better. After all, they guarantee no pod eviction. But recall that their restrictiveness makes it hard to perform maintenance on a cluster. What I would ideally like to achieve is to have the same level of certainty that a user's workloads won't be interrupted, while still allowing the host to maintain the cluster without headaches. We can try and roll the feature out under a flag and see how stable user's workloads will be. Until then, I wouldn't remove PDBs, even if I'm not a fan of them.

vpoliakov-pixel: Okay, let's keep them both and play around for now

PR #39, opened 2026-03-29
[JBRes-5153] Support arbitrary base image building

samorojy:
Resolves: JBRes-5153


samorojy:
Main thoughts / TODOs:
1. There should be a better way to handle the current mix of API. Probably we can try to add a separate method / discontinue some of the existing ones. 
2. Reiterate through plugin implementations / BuildContext. It might not be very clear to have reused "context values" through different plugins. 
3. Add tests for YAML/Dockerfile generation
4. tba

draft-demo/image_build_api_usage_draft.py
samorojy, line 1: that's an example

orchestrator/src/idegym/orchestrator/kaniko_docker_api.py
```diff
@@ -144,37 +106,18 @@ async def build_and_push_images(self, path: Path) -> List[str]:
...
-                if pod_resources is not None and type(pod_resources) is not dict:
-                    raise TypeError(f"Unable to parse 'resources' field. Expected a dictionary, got: {pod_resources}")
-                project = item.pop("project", {})
-                if type(project) is not dict:
-                    raise TypeError(f"Unable to parse 'project' field. Expected a dictionary, got: {project}")
-                model = GitRepositoryResource if "path" in project else GitRepositorySnapshot
+            pipeline = parse_image_build_pipeline(file.read())
+            for item in pipeline.images:
+                if item.request is None:
+                    raise ValueError("Kaniko builds require a project download request")
```
dabico: Maybe we can move this check to `validate_yaml_content` to fail at request time?

common-utils/src/idegym/image/plugins.py
```diff
@@ -0,0 +1,401 @@
...
+            "extract $IDEGYM_PROJECT_ARCHIVE_PATH $IDEGYM_PROJECT_ROOT",
+        ]
+        if owner is not None:
+            commands.append(f"chown -R {owner}:{group} {quote(ctx.project_root)}")
+
+        return _render_run_block(commands, comment="Fetch and unpack the project")
+
+
+@dataclass(frozen=True, slots=True)
+class IdegymServer(PluginBase):
```
dabico: Yet another naming scheme! 😅

common-utils/src/idegym/image/plugins.py
```diff
@@ -0,0 +1,401 @@
...
+        return ctx.updated(context_path=self.root)
+
+    def render(self, ctx: BuildContext) -> str:
+        if self.source == "git":
+            raise NotImplementedError("IdegymServer.from_git(...) is not implemented yet")
+
+        user = ctx.current_user
+        group = str(ctx.get_extra("idegym.user.group", user))
+
+        return "\n".join(
```
dabico: Why not use a multi-line string literal? It's both easier to read, and we can annotate it with the comment to use syntax highlighting in the IDE. Would make both the reading and writing much easier.

common-utils/src/idegym/image/plugins.py
```diff
@@ -0,0 +1,401 @@
...
+__all__ = [
+    "BuildContext",
+    "BaseSystem",
+    "IdegymServer",
+    "Permissions",
+    "Plugin",
+    "PluginBase",
+    "Project",
+    "User",
+]
```
dabico: Can we use an immutable collection for exports?


```suggestion
__all__ = (
    "BuildContext",
    "BaseSystem",
    "IdegymServer",
    "Permissions",
    "Plugin",
    "PluginBase",
    "Project",
    "User",
)
```

common-utils/src/idegym/image/plugins.py
```diff
@@ -0,0 +1,401 @@
...
+            "fi".format(
+                username=self.username,
+                uid=self.uid,
+                group=group,
+                home=quote(home),
+                shell=quote(self.shell),
+                add_groups=f"usermod -aG {additional_groups} {self.username}" if additional_groups else ":",
+                group_flags=group_flags,
+                create_home=create_home_flag,
+            )
```
dabico: Why use `format` instead of f-string?

common-utils/src/idegym/image/plugins.py
```diff
@@ -0,0 +1,401 @@
...
+            ),
+        ]
+
+        for extra_group in self.additional_groups:
+            commands.append(f"if ! getent group {extra_group} >/dev/null 2>&1; then groupadd {extra_group}; fi")
+
+        if additional_groups:
+            group_flags = f"-G {additional_groups}"
+        else:
+            group_flags = ""
```
dabico: Can be replaced with a simple ternary:


```suggestion
        group_flags = f"-G {additional_groups}" if additional_groups else ""
```

image-builder/src/idegym/image/plugins.py
```diff
@@ -0,0 +1,401 @@
...
+@dataclass(frozen=True, slots=True)
+class BaseSystem(PluginBase):
+    DEFAULT_PACKAGES: ClassVar[tuple[str, ...]] = (
+        "bash",
+        "ca-certificates",
+        "coreutils",
+        "curl",
+        "dumb-init",
+        "findutils",
+        "fontconfig",
```
dabico: I think this is some holdover package that was used for internal tooling. I don't think it's necessary anymore.

image-builder/src/idegym/image/plugins.py
dabico, line 1: Why use `dataclass` instead of Pydantic models?

common-utils/src/idegym/image/plugin.py
```diff
@@ -0,0 +1,45 @@
...
+@dataclass(frozen=True, slots=True)
+class BuildContext:
+    base: str
+    current_user: str = "appuser"
+    home: str = "/home/appuser"
+    project_root: str = "/home/appuser/work"
+    request: DownloadRequest | None = None
+    labels: Dict[str, str] = field(default_factory=dict)
+    context_path: str = "."
+    extras: Mapping[str, Any] = field(default_factory=dict)
```
dabico: `Dict` and `Mapping` have different semantics. The latter provides only a subset of methods that `Dict` does. What was the intent here? To signal that `labels` are mutable while `extras` are not?

common-utils/src/idegym/image/builder.py
```diff
@@ -0,0 +1,129 @@
+from dataclasses import dataclass, replace
+from pathlib import Path
+from typing import Any, Dict, Optional, Tuple
+
+from idegym.api.docker import BaseImage
+from idegym.api.image_build import ImageBuildPipeline, ImageBuildSpec
+
+from .plugin import BuildContext, Plugin
```
dabico: Please use absolute paths when importing from other modules or workspaces:


```suggestion
from idegym.image.plugin import BuildContext, Plugin
```

image-builder/src/idegym/image/plugin.py
```diff
@@ -0,0 +1,45 @@
...
+
+from idegym.api.download import DownloadRequest
+
+
+@dataclass(frozen=True, slots=True)
+class BuildContext:
+    base: str
+    current_user: str = "appuser"
+    home: str = "/home/appuser"
+    project_root: str = "/home/appuser/work"
```
dabico: Should the project location be in the user home directory by default? I've seen other projects use a dedicated root-level directory owned by the user, such as `/workspace`.

api/src/idegym/api/orchestrator/build.py
```diff
@@ -13,9 +12,9 @@ class BuildFromYamlRequest(BaseModel):
     @field_validator("yaml_content")
     def validate_yaml_content(cls, value: str) -> str:
         try:
-            parse_yaml(value)
+            parse_image_build_pipeline(value)
             return value
-        except YAMLError as ex:
+        except Exception as ex:
```
dabico: So `parse_image_build_pipeline` can throw either:

- `TypeError` from what you defined
- `YAMLError` from `yaml.safe_load`
- `ValidationError` from Pydantic's `model_validate`

I'd recommend avoiding such generic `Exception` matching. For the first two, you can indeed catch them and re-throw:

```suggestion
        except TypeError, YAMLError as ex:
```

And let `ValidationError` propagate naturally.

api/src/idegym/api/image_build.py
```diff
@@ -0,0 +1,76 @@
...
+        return self.model_dump(mode="json", exclude_none=True)
+
+    def image_version(self) -> str:
+        digest = md5()
+        if self.request is not None:
+            digest.update(dump_json(self.request.descriptor.model_dump(mode="json"), sort_keys=True).encode())
+        digest.update(dump_json(self.labels, sort_keys=True).encode())
+        digest.update(self.context_path.encode())
+        digest.update(self.dockerfile_content.encode())
+        return digest.hexdigest()
```
dabico: In #35 I introduced a utility function:

https://github.com/JetBrains-Research/idegym/blob/bfecb7b81cc504c0ac4f0b0f0c9442ecc3d96128/common-utils/src/idegym/utils/hashing.py#L1-L9

Can you use that here as well?

api/src/idegym/api/orchestrator/build.py
dabico: More on [Pydantic validation errors](https://docs.pydantic.dev/latest/errors/errors).

dabico (dismissed):
I'll do another round later. I'd like to play with the code a bit and get a feel for it. For now, I've left comments about issues I've noticed.

common-utils/src/idegym/image/plugins.py
samorojy: There was an idea with this approach -- to generate content in place easily: `*(f"COPY --chown={user}:{group} {path} {path}/" for path in self.WORKSPACE_DIRS)`

image-builder/src/idegym/image/plugins.py
samorojy: Thanks! I think we need to go through the whole Dockerfile and remove outdated things. But I'd vote it to be the latest step as it is easier to handle :)

image-builder/src/idegym/image/plugin.py
samorojy: The same here. 🙏🏻

api/src/idegym/api/image_build.py
samorojy: Thanks, will merge main and reuse it here.

samorojy:
The next steps:
- Fix artefacts like wrong typing / Pydantic models fields 
~~- Move everything related to DockerService / Image Building into a separate workspace / module.~~
- Pass image definition YAML to orchestrator and render dockerfile from it for kaniko
- Polish plugins implementation (include deciding how updated dockerfile should look) 
- Tests, Tests, Tests!

image-builder/src/idegym/image/plugin.py
vpoliakov-pixel: less problems with ownership and access, lets keep it here for now

vpoliakov-pixel (approved):

PR #49, opened 2026-04-02
[JBRes-8914] E2E tests in CI

dabico:
Launch E2E tests on CI with larger runners. Remove the lock file from the E2E test directory, as only one should be used in the root. Rename test directory to `e2e-tests` to match existing conventions and simplify mapping to markers. Flattened package structure to match other test suites. Remove workflow steps that provided questionable benefits while slowing each test suite down for 5+ minutes.

Resolves: JBRes-8914


.github/actions/prepare-docker-build/action.yml
```diff
@@ -23,5 +23,5 @@ runs:
       uses: docker/login-action@v3
       with:
         registry: ghcr.io
-        username: ${{ github.actor }}
+        username: github-actions[bot]
```
dabico, line 26: I noticed that _I_, the actor who triggered the action, would be logged in. I just switched it to a neutral actor.

.github/workflows/ci.yml
```diff
@@ -46,14 +50,31 @@ jobs:
     steps:
       - name: Checkout latest
         uses: actions/checkout@v6
-      - name: Remove Unwanted Software
-        uses: ./.github/actions/remove-unwanted-software
       - name: Install Docker
+        if: matrix.suite != 'e2e'
```
dabico, line 54: So I didn't know this, but all GitHub Actions runners come with Docker pre-installed. The `docker/setup-docker-action` doesn't install Docker from scratch, it just installs a specific version with its own daemon, socket, and context. We skip the setup for E2E because that separate Docker context was the root cause of multiple issues with `minikube`. Either `minikube image load` would fail or we'd get errors like "No such container: minikube". The E2E tests work fine with the runner's pre-installed Docker because `docker build` works (we don't need anything fancy like multi-platform builds), `minikube image load` works and there are no context conflicts. The integration tests use this action because they need the `network=host` driver option so the containerized builder can push to the registry service at `localhost:5000`. Whether they'd work without, and whether we still need it, are open questions.

.github/workflows/ci.yml
```diff
@@ -65,12 +86,19 @@ jobs:
...
         run: scripts/build_server_images.py --skip-base ubuntu
       - name: Install dependencies
         run: uv sync --all-packages --all-extras --all-groups
       - name: Run ${{ matrix.suite }} tests
-        run: uv run pytest -vv -m "${{ matrix.suite }}"
+        run: |
+          if [ "${{ matrix.suite }}" = "e2e" ]; then
+            sudo -E kubectl port-forward -n ingress-nginx svc/ingress-nginx-controller 80:80 > /dev/null 2>&1 &
+            sleep 5
+          fi
```
dabico, line 101: What `minikube tunnel` does is it creates network routes and assigns external IPs to `LoadBalancer` services. Since the enterprise runners have "Public IP: Disabled", this means that the tunnel process runs fine, but there's no external IP to assign. The tunnel is designed for environments where the host can receive routable traffic. On the other hand, `kubectl port-forward` works at a different level: it creates a direct TCP tunnel from `localhost:80` on the runner to port 80 on the ingress controller pod, entirely within the Kubernetes API. It doesn't need external IPs, network routes, or public networking. It's just a pipe through the Kubernetes API server, which works regardless of the host's network configuration.

dabico: Furthermore, the Kubernetes API client we use should have port-forwarding capabilities. We should investigate if it's possible to avoid having to create manual tunnels in a separate terminal. I did this when creating the first local deployment guide, but only because it felt more natural at the time. I wonder if we can completely forego modifying `/etc/hosts` and tunneling in a separate terminal window in favor of programmatically creating a port-forward once and using `localhost:80` as the orchestrator URL.

e2e-tests/utils/k8s_setup.py
```diff
@@ -13,7 +13,9 @@
 from utils import k8s_client
 from utils.constants import (
     BASE_URL,
+    DEFAULT_HEALTH_CHECK_TIMEOUT,
     DEFAULT_NAMESPACE,
+    HEALTH_CHECK_INTERVAL,
```
dabico, line 18: These were unused, so they seemed perfect for the utilities.

.pre-commit-config.yaml
```diff
@@ -13,10 +13,11 @@ repos:
     - id: name-tests-test
       args: [ "--pytest-test-first" ]
       files: ^.*-tests/.*\.py$
+      exclude: ^.*-tests/(utils|config)/
```
dabico, line 16: We didn't use packages in other test suites, so we never had problems with module names that didn't have the `test` prefix. Changing the name of the folder helped reveal this issue. Let's relax the check to allow some kind of utility packages within tests.

.github/workflows/ci.yml
```diff
@@ -26,18 +26,22 @@ jobs:
...
         uses: astral-sh/ruff-action@v3
         with:
-          version: "0.15.4"
+          version: "0.15.8"
           args: "check --output-format github"
 
   test:
     name: Run ${{ matrix.suite }} tests
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
```
dabico, line 34: This is debatable. From what I've seen, the usual completion order is `unit` then `integration` then `e2e`. Failure in one of these earlier stages usually resulted in a guaranteed failure in later ones. I think it's fine if we revert to the old behavior of cancelling the test suite chain if one of the earlier suites fails.

.github/workflows/ci.yml
vpoliakov-pixel: In the upcoming pull request there are some kaniko tests, so registry service thing could be useful. But probably everything will work with the default docker

.github/workflows/ci.yml
vpoliakov-pixel: Direct port forwarding looks better for tests but for local deployment it will still be some sort of tunnelling, I guess.
Using port forwarding through code seems to be great too.

.pre-commit-config.yaml
vpoliakov-pixel: Hardcoding utils and config here looks strange to me.
Why not use packages and not flat structure?

.github/workflows/ci.yml
vpoliakov-pixel: Fail fast is great but it could be better to collect all available errors in one go. Compilation errors should be caught locally before running cloud ci (meaning that running unit and integration tests could be even mandatory before pushing)

vpoliakov-pixel (approved):

vpoliakov-pixel:
I have some concerns about making back and forth e2e tests a workspace but I do not have a strong opinion here so lets keep it like this for now. Maybe in future we will put all tests in one package/workspace with 3 sub folders of some sort

PR #63, opened 2026-04-14
[JBRes-8561] review http error codes, document them and fix some issues related to it

vpoliakov-pixel:
Resolves: JBRes-8561


dabico (approved):

server/router/fs.py
```diff
@@ -110,5 +110,10 @@ async def rm(
     elif recursive:
         await rmtree(path)
     else:
+        if [_ async for _ in path.iterdir()]:
```
samorojy, line 113: a bit weird to read that syntax but ok since it's short. 

We tend to iterate over the whole dir instead of stopping on the first match. This might not be quite what we were aiming for :)

samorojy (commented):
LGTM!

PR #69, opened 2026-04-16
[JBRes-8920] Support for dedicated node pools

dabico:
Resolves: JBRes-8920

### DIFF
--- a/api/src/idegym/api/config.py
+++ b/api/src/idegym/api/config.py
@@ -81,6 +81,14 @@ class AsyncioConfig(BaseModel):
     dump_interval: int = Field(description="Interval in seconds between asyncio task dumps", ge=1, default=300)
 
 
+class NodePoolConfig(BaseModel):
+    enabled: bool = Field(description="Enable dedicated node pool scheduling", default=False)
+    taint_key: str = Field(description="Taint key applied to dedicated pool nodes", default="jetbrains.com/idegym")
+    preference_weight: int = Field(
+        description="Weight (1-100) for preferring dedicated pool nodes", ge=1, le=100, default=100
+    )
+
+
 class ResourcesConfig(BaseModel):
     default_cpu_request: float = Field(description="Default CPU cores per environment", ge=0, default=1.0)
     default_ram_request: float = Field(description="Default RAM per environment in GB", ge=0, default=2.0)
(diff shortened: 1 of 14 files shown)


api/src/idegym/api/config.py
```diff
@@ -81,6 +81,14 @@ class AsyncioConfig(BaseModel):
     dump_interval: int = Field(description="Interval in seconds between asyncio task dumps", ge=1, default=300)
 
 
+class NodePoolConfig(BaseModel):
+    enabled: bool = Field(description="Enable dedicated node pool scheduling", default=False)
+    taint_key: str = Field(description="Taint key applied to dedicated pool nodes", default="jetbrains.com/idegym")
```
vpoliakov-pixel, line 86: Do we need `jetbrains.com/` prefix?

orchestrator/kubernetes/grafana/deployment.yaml
```diff
@@ -19,6 +19,18 @@ spec:
...
+        - key: jetbrains.com/idegym
+          operator: Exists
+          effect: NoSchedule
+      affinity:
+        nodeAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - weight: 100
+              preference:
+                matchExpressions:
+                  - key: jetbrains.com/idegym
```
vpoliakov-pixel, line 32: Probably we should not allow to configure this value then if we hardcode it here.
Or it should be explicitly documented at least.

vpoliakov-pixel (approved):

vpoliakov-pixel:
Would be great to document this feature explicitly somewhere.
Maybe by adding more info around config feature flag description.

api/src/idegym/api/config.py
dabico: Wouldn't say it's strictly "needed" but it's a good practice to prefix domain-specific annotations/labels/taints, as it more clearly conveys where they originate from.

orchestrator/kubernetes/grafana/deployment.yaml
dabico: Hmmm yeah, that's a good point. Making it configurable via Helm values might be a better option. But we currently lack a chart for the orchestrator. What should we do for the time being? Hard-coding it would simplify code at the cost of configurability. At the same time, I'm not really sure if there are use cases where a client would want to customize the label. Should we hard-code it for now and return to this once we get clearer demands?

vpoliakov-pixel: Lets keep it as is and write a comment near config files / documentation

vpoliakov-pixel (approved):

PR #85, opened 2026-04-22
[JBRes-9108] github issue templates for bug report and feature request

iamblobfish:
Adding github issue templates for feature request and bug report.

Resolves: JBRes-9108

PR #86, opened 2026-04-22
[JBRes-9108]  github issue templates

iamblobfish:
Added two issue templates: feature request and bug report
Closes: #85


iamblobfish:
image
image

dabico:
> ### Optional additional items
>
> #### Assignees:

Does this mean that the issue submitter sets the assignee? Shouldn't this be handled on our end?

dabico:
Also, there are two:

> #### Alternatives Considered

Sections in the feature proposal

iamblobfish:
> Does this mean that the issue submitter sets the assignee? Shouldn't this be handled on our end?

I think that means that we could set default assignee, but I didn't that is why the field is blank. It is default field, it couldn't be turned off.
image

vpoliakov-pixel (approved):

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,29 @@
...
+
+**Proposed Solution**
+
+How would you like this to work? Include any API shape, config format, or behavior you have in mind.
+
+**Alternatives Considered**
+
+Any other approaches you've thought about and why you ruled them out.
+
+**Alternatives Considered**
```
vpoliakov-pixel: Doublicate

iamblobfish: fixed

PR #91, opened 2026-04-29
[JBRes-9069] Introducing plugins for tools and rewards

vpoliakov-pixel:
Initial version of pluggable architecture for image building, endpoints/mcp declarations in idegym server and client methods

Resolves: JBRes-9069


plugins/defaults/src/idegym/plugins/defaults/image.py
vfedorishchev, line 682:
```suggestion
                cp -r /tmp/idegym-src/rewards $IDEGYM_PATH/rewards; \\
                cp -r /tmp/idegym-src/plugins $IDEGYM_PATH/plugins; \\
```
are we missing another instruction here, like:
`cp -r /tmp/idegym-src/plugins $IDEGYM_PATH/plugins; \\`

and then the test can be extended with

```def test_idegym_server_from_git_render_copies_plugins_workspace():
    plugin = IdeGYMServer.from_git(url="https://github.com/owner/idegym.git")
    ctx = BuildContext(base="debian:bookworm-slim")
    fragment = plugin.render(ctx)
    assert "cp -r /tmp/idegym-src/plugins $IDEGYM_PATH/plugins" in fragment```

File: documentation/client.md [hunk 1]
```diff
@@ -236,6 +236,69 @@ response = await server.restart_server(
...
+
+The `pycharm` attribute is only present if the `idegym-plugin-pycharm` package is installed and
+the `pycharm` client entry point loads successfully. Accessing it on a server whose image was built
+without the PyCharm plugin will raise `AttributeError`.
+
+### Checking for a plugin
+
+```python
+if hasattr(server, "pycharm"):
+    health = await server.pycharm.health()
```
vfedorishchev, line 294: Maybe worth tweaking this wording/example a bit: `server.pycharm` is attached from the locally installed `idegym.plugins.client` entry point, so `hasattr(server, "pycharm")` only tells us the client wrapper loaded. It doesn't necessarily mean the running server image has the PyCharm server plugin enabled; in that case `server.pycharm.health()` may still hit
  `/api/pycharm/health` and get a 404.

  Could we either document that distinction here, or add a runtime capability/health check before presenting this as the recommended way to check remote plugin support?

documentation/plugins.md
```diff
@@ -0,0 +1,570 @@
...
+
+The `IdeGYMServer` plugin always includes `tools` and `rewards`. Additional plugins can be added
+by other `@image_plugin` classes that call `ctx.with_extra("idegym.enabled_server_plugins", [...])`.
+The `PyCharm` plugin does this automatically when added to an image:
+
+```python
+image = (
+    image
+    .with_plugin(PyCharm(version="2024.3.1", user="appuser"))   # adds "pycharm" to the list
+    .with_plugin(IdeGYMServer.from_local(root=from_root()))      # writes plugins.json
```
samorojy, line 393: didn't get the comments here 😔

plugins/pycharm/src/idegym/plugins/pycharm/server.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+
+    _MCP_URL = "http://localhost:6789/mcp"
+
+    @classmethod
+    def get_server_router(cls):
+        try:
+            from fastapi import APIRouter
+        except ImportError:
+            return None
```
samorojy, line 26: Is it a case?

samorojy (approved):
Tbh, I got overwhelmed by the number of changes. 
My suggestion is to merge it after small fixes mentioned above and then collect feedback, because I have a feeling that it could be overcomplicated for some users.

documentation/plugins.md
vpoliakov-pixel: """Embed the IdeGYM server into the image.

    Installs the server runtime, supervisord configuration, entrypoint scripts, and sets
    up the container's ``CMD`` / ``HEALTHCHECK``. This plugin must be the last one in the
    pipeline as it emits the full container entrypoint.

    Also writes ``/etc/idegym/plugins.json`` listing the server plugins to enable at
    runtime. Built-in plugins (``tools``, ``rewards``) are always included. Optional
    plugins that called ``apply()`` earlier (e.g. ``PyCharm``) are read from
    ``ctx.extras["idegym.enabled_server_plugins"]``.

    Use ``from_local()`` to copy from a local workspace (the build context is set to
    ``root``), or ``from_git()`` to clone from a remote repository inside the container.
    """

plugins/pycharm/src/idegym/plugins/pycharm/server.py
vpoliakov-pixel: It allows to use the whole plugin package as one piece in places where you do not fastapi (image builder and client)

plugins/defaults/src/idegym/plugins/defaults/image.py
vpoliakov-pixel: You are correct, thanks, from_local also should have plugins copied
Fixed in the next PR https://github.com/JetBrains-Research/idegym/pull/93

documentation/client.md
vpoliakov-pixel: Also great idea, I've introduced `capabilities` endpoint that reports all available plugins on the server
New functionality and appropriate docs are here https://github.com/JetBrains-Research/idegym/pull/93

vfedorishchev (approved):

PR #92, opened 2026-04-29
[JBRes-9089] Introduce API for GKE pod checkpoint

samorojy:
Resolves: JBRes-9089


vpoliakov-pixel:
It would be great to be able to ask to make a snapshot from the client.
Introducing a specific class or maybe it is already available in some library for pod snapshot CRD also makes sense

samorojy:
Since these CRDs are from third party provider (GCP), I doubt that they are in some library. 
Thus, we can wrap them into Pydantic models and this would be a nice improvement for the next iteration. (JBRes-9179)

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -635,19 +639,20 @@ async def check_and_delete(
...
     """
-    Delete the Deployment for a server (Service and PDB are garbage-collected via owner references).
+    Delete the Deployment for a server.
 
     Raises ResourceDeletionFailedException if the Deployment cannot be deleted.
     """
-    async with async_kube_api() as (apps, _, _, _):
-        if not await check_and_delete(
+    async with async_kube_api() as (apps, _, _, _, _):
+        deployment_deleted = await check_and_delete(
```
dabico, line 650: What's with these changes? Was any of the functionalities mentioned in the comment changed? Don't see it among the diffs.

vpoliakov-pixel (approved):

orchestrator/src/idegym/orchestrator/pod_snapshot.py
```diff
@@ -0,0 +1,76 @@
...
+from idegym.backend.utils.kubernetes_client import async_kube_api
+from idegym.utils.logging import get_logger
+
+logger = get_logger(__name__)
+
+CRD_GROUP = "podsnapshot.gke.io"
+CRD_VERSION = "v1"
+
+
+class PodSnapshotService:
```
dabico, line 11: Why introduce a whole abstraction instead of just helper functions? As far as I can tell, `_config` isn't used and `_namespace` can just be a function parameter.

orchestrator/src/idegym/orchestrator/router/snapshot.py
```diff
@@ -0,0 +1,124 @@
...
+    config: Config = low_level_request.app.state.config
+    snapshot_config = config.orchestrator.pod_snapshot
+
+    if not snapshot_config.enabled:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Pod snapshot feature is not enabled",
+        )
+
+    server = await validate_server(client_id=request.client_id, server_id=request.server_id)
```
dabico, line 37: According to [limitations](https://docs.cloud.google.com/kubernetes-engine/docs/concepts/pod-snapshots#limitations), this feature only works with `gvisor` container runtime. I think we need a different kind of "validate" utility that checks preconditions before proceeding with snapshotting.

dabico (changes requested):
Apart from raised concerns, I have a question: should we also create a "preliminary" snapshot of an environment right after it enters "Ready" state? This way, users have an initial checkpoint they can roll back to, without having to use the API explicitly.

orchestrator/src/idegym/orchestrator/router/snapshot.py
dabico: Equally important:

> Pod snapshots don't support E2 machine types.

orchestrator/src/idegym/orchestrator/pod_snapshot.py
dabico: Also, `backend-utils` might be a better place for this code, since this is where all other Kubernetes utilities reside.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
samorojy: I extended create_clients by CustomObjectsApi, so thats why we need to properly unpack it here: `async with async_kube_api() as (apps, _, _, _, _)`

changes in `if condition' to be used later, when we will worry about cleaning up Triggers Crd

orchestrator/src/idegym/orchestrator/pod_snapshot.py
samorojy: The initial assumption was that we need to create Storage and Snapshot policies per pod. So, that's why it seemed a better idea to encapsulate the logic inside the class.  

There are still a lot of uncertainties about the logic behind and GCP this implementation, and I doubt that this is the final solution. For now we want to achieve PoC.

orchestrator/src/idegym/orchestrator/router/snapshot.py
samorojy: That's not the goal of the PoC.

samorojy:
> I have a question: should we also create a "preliminary" snapshot of an environment right after it enters "Ready" state? This way, users have an initial checkpoint they can roll back to, without having to use the API explicitly.

Define `Ready state`. I think that it is up to the user to decide what their start point is. We have discussed this point with Vladimir, so please check [JBRes-9180](https://youtrack.jetbrains.com/issue/JBRes-9180) for a rough sketch.

client/src/idegym/client/operations/servers.py
```diff
@@ -195,3 +196,24 @@ async def finish_server(
         request = FinishServerRequest(client_id=client_id, namespace=namespace, server_id=server_id)
         response_raw = await self._utils.make_request("POST", "/api/idegym-servers/finish", request)
         return ServerActionResponse.model_validate(response_raw)
+
+    async def snapshot_server(
+        self,
+        server_id: int,
```
vfedorishchev, line 197: I was wondering if we have an agreement to avoid API objects to be forwarded deeper in the code base, since this one is practically the same as `CreateSnapshotRequest` + polling_config

dabico (approved):

orchestrator/src/idegym/orchestrator/pod_snapshot.py
vfedorishchev: shall we discuss this in a tech sync call. I think our current choices between primitives and objects, returning tuples vs objects, as well as OOP abstractions and procedural programming practices aren't well agreed on.

orchestrator/src/idegym/orchestrator/router/snapshot.py
samorojy: [JBRes-9184](https://youtrack.jetbrains.com/issue/JBRes-9184) 

Good note! Thanks.

api/src/idegym/api/orchestrator/snapshots.py
```diff
@@ -0,0 +1,17 @@
+from typing import Optional
+from uuid import UUID
+
+from pydantic import BaseModel, Field
+
+
+class CreateSnapshotRequest(BaseModel):
```
vfedorishchev, line 7: I'm working on the introduction of the MCP server support on the Orchestrator level, which would wrap around the Orchestrator API.
It would really help if you can add the description fields for the API classes, to help agents understand the purpose of each field a bit better.
You can check my PR #90 if you need an example

client/src/idegym/client/operations/servers.py
samorojy: > I was wondering if we have an agreement to avoid API objects to be forwarded deeper in the code base

If I understood your question correctly, then yes. This implementation is done in sync with how we currently pass arguments around. 

We can discuss this further on the technical discussion sync

orchestrator/src/idegym/orchestrator/pod_snapshot.py
samorojy: Let's discuss it.
I think It is under Minor priority rn, wdyt?

PR #93, opened 2026-05-02
[JBRes-9072] introduce full pycharm and idea plugins

vpoliakov-pixel:
- plugins for having pycharm and idea as mcp servers over the internal project and for providing independent inspect.sh script integration for inspections without running full IDE
- complete e2e tests for both plugins skipped in ci integrations because they are long and download a lot of things from the internet

Resolves: JBRes-9072


api/src/idegym/api/plugin.py
```diff
@@ -93,14 +93,30 @@ def render(self, ctx: BuildContext) -> str:
...
+        Each returned string is a complete ``FROM … AS …`` stage. Stages are inserted
+        before the primary ``FROM`` instruction so they can produce artifacts that are
+        later consumed via ``COPY --from=<stage>``.
+
+        Override when your plugin needs to compile an artifact (e.g. a JetBrains plugin
+        JAR) inside Docker without including the build toolchain in the final image.
+        """
+        return []
+
+    def get_mcp_upstream(self, ctx: "BuildContext") -> Optional[str]:
```
vfedorishchev: is it meant to be a string or the instance of BuildContext?

client/src/idegym/client/operations/servers.py
```diff
@@ -192,6 +193,14 @@ async def finish_server(
         response_raw = await self._utils.make_request("POST", "/api/idegym-servers/finish", request)
         return ServerActionResponse.model_validate(response_raw)
 
+    async def capabilities(self, server_id: int, client_id: Optional[UUID] = None) -> CapabilitiesResponse:
```
vfedorishchev: nit, prefix this one with a verb, unless we want to define it as a param, the same comment as for server wrapper on the client side

e2e-tests/resources/mcp_wait_180s.sh
```diff
@@ -0,0 +1,19 @@
+# Poll the MCP SSE endpoint every 3s for up to 180s (60 iterations).
+# Exits 0 with "SUCCESS" on the first 200 response; exits 1 after timeout.
+# Used by IDEA and PyCharm inspect/plugin tests that share the same 180s budget.
+for i in $(seq 1 60); do
+    http_code=$(curl -s -o /dev/null -w "%{http_code}" --max-time 3 \
+        "http://localhost:64342/sse" 2>/dev/null || true)
```
vfedorishchev, line 6: I looked through the [IntelliJ repository](https://github.com/JetBrains/intellij-community/blob/81c743b09802fa039aa1cc87e351be3dcc9122a7/build/mcp-servers/ij-proxy/ij-mcp-proxy.ts) to figure out if there is a better way of checking MCP server availability and I've got a few questions.

1. It seems like a user need to follow certain setup steps to enable the MCP server on the IDE side. Have those been considered by us?
2. You make use of `/sse` but it looks like they've changed it to `/stream`, have you tested if `/sse` endpoint is functional?
3. is `/stream` a good indicator that the MCP server started and running well, or do we need to at least call `list_tools` endpoint to confirm the server is operational?

plugins/defaults/src/idegym/plugins/defaults/inspect.py
```diff
@@ -0,0 +1,113 @@
...
+        cmd.extend(["-d", request.directory])
+    cmd.extend(["-format", request.format])
+    cmd.append(f"-v{request.verbosity}")
+
+    proc = await asyncio.create_subprocess_exec(
+        *cmd,
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.STDOUT,
+    )
+    await asyncio.wait_for(proc.communicate(), timeout=request.timeout)
```
vfedorishchev: I think we might need to kill the process after the operation timed out.

image-builder/src/idegym/image/builder.py
```diff
@@ -212,8 +220,13 @@ def _render_dockerfile(
         self,
         ctx: BuildContext,
         fragments: tuple[str, ...],
+        build_stages: tuple[str, ...] = (),
```
vfedorishchev: does it have to be a tuple? If not and we are fine with the list maybe we can split this method into smaller ones and use the list unpack? Not critical but can better express the logic behind each step:

```
def _render_dockerfile(
      self,
      ctx: BuildContext,
      fragments: tuple[str, ...],
      build_stages: tuple[str, ...] = (),
  ) -> str:
      sections = [
          *build_stages,
          self._render_base_stage_header(),
          self._render_project_archive_env() if ctx.request is not None else "",
          f'ENV IDEGYM_PROJECT_ROOT="{ctx.project_root}"',
          *fragments,
          f"USER {ctx.current_user}",
          _run_block(self.commands),
      ]

      return "\n\n".join(section for section in sections if section.strip()).strip() + "\n"

  With helpers like:

  def _render_base_stage_header(self) -> str:
      return "\n\n".join(
          [
              f"FROM {self.base}",
              'SHELL ["/bin/bash", "-c"]',
              "USER root",
          ]
      )


  def _render_project_archive_env(self) -> str:
      return "\n".join(
          [
              "ARG IDEGYM_PROJECT_ARCHIVE_URL",
              "ARG IDEGYM_PROJECT_ARCHIVE_PATH",
              "ARG IDEGYM_AUTH_TOKEN",
              "ARG IDEGYM_AUTH_TYPE",
              "",
              'ENV IDEGYM_PROJECT_ARCHIVE_URL="$IDEGYM_PROJECT_ARCHIVE_URL"',
              'ENV IDEGYM_PROJECT_ARCHIVE_PATH="$IDEGYM_PROJECT_ARCHIVE_PATH"',
          ]
      )

```

vfedorishchev (approved):

api/src/idegym/api/plugin.py
vpoliakov-pixel: Yes, it should return an mcp url to connect, that then could be placed into `/etc/idegym/mcp-upstreams.d/{plugin_name}.json` during build stage

BuildContext is needed because mcp server (through mcp server plugin) won't work without opened project

client/src/idegym/client/operations/servers.py
vpoliakov-pixel: fixed

plugins/defaults/src/idegym/plugins/defaults/inspect.py
vpoliakov-pixel: fixed

image-builder/src/idegym/image/builder.py
vpoliakov-pixel: fixed

e2e-tests/resources/mcp_wait_180s.sh
vpoliakov-pixel: 1. Everything is covered by us - we open project on startup using a plugin, we put MCP server enabling config and we download and put a specific version of MCP server plugin to be loaded by IDE
2. Yes, /sse is working and functional otherwise what is the point of the PR. But I agree that we need to switch to stream since it is a common practice now. Also we can force using the latest versions of IDEs - the mcp plugin is bundled now.

vpoliakov-pixel: 3. For /sse the current approach worked well, not sure how should we tackle the /stream. /list_tools looks like an overkill for that

api/src/idegym/api/plugin.py
vpoliakov-pixel: fixed, now it is get_mcp_upstream(self, ctx: BuildContext)

e2e-tests/resources/mcp_wait_180s.sh
vfedorishchev: as long as /sse is a valid endpoint I'm fine with the current state

PR #97, opened 2026-05-04
[JBRes-8916] Helm chart

dabico:
### Summary

**Create the official project Helm chart,** migrating the service deployment from raw manifests + Kustomize. Furthermore, usage of the chart was wired directly into E2E tests. Details won't be covered in the summary, since changes have been mirrored to the [documentation](https://github.com/JetBrains-Research/idegym/tree/bd49ad62ea064d4556e9c61d745704c358eeba23/documentation).

### Key changes

- **Created Helm chart in [charts/idegym](https://github.com/JetBrains-Research/idegym/tree/bd49ad62ea064d4556e9c61d745704c358eeba23/charts/idegym)**: templated all existing resources from static manifests such as deployments, ingresses, RBAC, services, and so on. Observability services and the database are now included as third-party Helm dependencies in the chart. In essence, we retained a simple deployment process while improving customizability and maintainability.
- **E2E tests on Helm**: `e2e-tests` now installs the chart via a dedicated [`values.yaml`](https://github.com/JetBrains-Research/idegym/blob/bd49ad62ea064d4556e9c61d745704c358eeba23/e2e-tests/config/values.yaml), instead of [`kustomization.yaml`](https://github.com/JetBrains-Research/idegym/blob/6171d6fe8e436b07b6faf3d18bc35f6315313526/e2e-tests/config/kustomization.yaml). Test scaffolding has been updated accordingly.
- **Removed static manifests**: deleted [orchestrator/kubernetes](https://github.com/JetBrains-Research/idegym/tree/6171d6fe8e436b07b6faf3d18bc35f6315313526/orchestrator/kubernetes) and the Kustomize overlays, making the chart the single source of truth.
- **Updated GitHub Actions workflows**:
    - Separate linting jobs into their own workflow file: [lint.yaml](https://github.com/JetBrains-Research/idegym/blob/4c50746604c39c464131f7ff9af77d3b8a1ee3cb/.github/workflows/lint.yml). Added a new job for linting Helm chart in CI.
    - Added a new chart [job](https://github.com/JetBrains-Research/idegym/blob/bd49ad62ea064d4556e9c61d745704c358eeba23/.github/workflows/publish.yml#L84-L111) in `publish.yml` for packaging and release.
    - Re-configured Dependabot for chart dependencies and orchestrator (now only scans `Dockerfile`).
- **Added `pre-commit` hook** for linting the chart. 
- **Updated documentation** with new instructions for E2E testing, local and remote deployment.

Resolves: JBRes-8916


vpoliakov-pixel:
Would it be reasonable to use pyhelm or smth like in e2e tests instead of running helm explicitly?

e2e-tests/conftest.py
```diff
@@ -97,12 +95,6 @@ def pytest_addoption(parser):
         default=False,
         help="Delete the entire idegym-local namespace after all tests complete",
     )
-    parser.addoption(
```
vpoliakov-pixel, line 100: I do not see any replacement of this functionality that could be useful, or should we stick to recreating the namespace in this case?

documentation/local_deployment.md
```diff
@@ -126,39 +129,125 @@ kubectl create secret docker-registry regcred \
...
+The chart deliberately does not manage credentials, expecting a set of pre-existing secrets in the release namespace.
+Only `postgres` is required for a minimal installation, while others are needed when you opt into observability services.
+
+#### `postgres` (required)
+
+The orchestrator and the bundled Bitnami PostgreSQL chart both read from a single `postgres` secret.
+You can create it like so:
+
+```shell
+kubectl create secret generic postgres -n idegym \
```
vpoliakov-pixel: Is it possible to provide a default user/password for at least local deployment? Extra configuration steps especially with values that you need to generate on your own could be tiring and not user friendly

vpoliakov-pixel (approved):

documentation/local_deployment.md
dabico: Simplified the process. `Secret` for Grafana and Postgres is now auto-generated. Included functionality and instructions for how they can be overridden with new values or secret references.

e2e-tests/conftest.py
dabico: Restored the beloved functionality

vpoliakov-pixel (approved):

dabico:
> Would it be reasonable to use pyhelm or smth like in e2e tests instead of running helm explicitly?

Forgot to address this, but [`pyhelm`](https://pypi.org/project/pyhelm/#history) seems like a dead project to me: no releases for close to 7 years. It doesn't even support Helm 3. I wouldn't recommend using it. If we are to use some kind of `pytest` utilities for Helm, then we should look for something more recent, or well-maintained. I haven't explored this space much. Considering our usage of Helm in E2E tests amounts to (un)installing releases, I think we can live with the subprocess approach for the time being.

PR #102, opened 2026-05-11
[JBRes-9332] Showing MCP tools from IdeGYM servers through orchestrator's mcp server

vpoliakov-pixel:
Resolves: JBRes-9332


DaniilMironovJB (approved):
🚀

PR #119, opened 2026-05-21
[JBRes-9180] initial implementation for checkpoint preparartion pipeline

vpoliakov-pixel:
Resolves: JBRes-9180


integration-tests/database/conftest.py
```diff
@@ -54,7 +54,9 @@ async def db(db_url: str):
     async with engine.begin() as conn:
         await conn.execute(
             text(
-                "TRUNCATE TABLE async_operations, job_statuses, servers, resource_limit_rules, clients"
+                # snapshot_jobs references snapshots and snapshot_prepare_requests, so listed first
```
samorojy: [No action needed]


afaik the order is only important for `DELETE` while the `CASCADE` should work fine with any order as it is an atomic operation. 

Probably just `snapshots` would be fine anyway, but for clarity let's leave as is.

integration-tests/docker_tests/test-project/.idea/misc.xml
samorojy, line 1: ?

orchestrator/src/idegym/orchestrator/snapshot_pipeline.py
```diff
@@ -0,0 +1,185 @@
...
+            wait_timeout=request.server_start_wait_timeout_in_seconds,
+        )
+
+        await update_server_status(server_id=server_id, availability_status=AvailabilityStatus.ALIVE)
+
+        snapshot_service = PodSnapshotService(config=pod_snapshot, namespace=request.namespace)
+        await snapshot_service.snapshot_server(server_name=server_generated_name)
+
+        # Brief pause to let the snapshot trigger propagate before stopping the pod.
+        await sleep(2)
```
samorojy: Awaiting for the snapshot is covered in implementation from #118

orchestrator/src/idegym/orchestrator/snapshot_pipeline.py
```diff
@@ -0,0 +1,185 @@
...
+
+        await _cleanup_server(
+            server_id=server_id, server_generated_name=server_generated_name, namespace=request.namespace
+        )
+
+        request_hash = compute_hash_for_start_request(request)
+        snapshot = await create_snapshot(
+            snapshot_name=str(server_id),
```
samorojy: It has been changed to server_generated name as of #118

samorojy: Probably its better to have 1 source of truth of what snapshot_id is. 
So, to return it from snapshot_server and propagate everywhere

orchestrator/src/idegym/orchestrator/snapshot_pipeline.py
```diff
@@ -0,0 +1,185 @@
...
+            environment_variables=(),
+            server_kind=request.server_kind,
+            snapshot_id=str(server_id),
+        )
+
+        await wait_for_pods_ready(
+            label_selector=f"app={server_generated_name}",
+            namespace=request.namespace,
+            wait_timeout=request.server_start_wait_timeout_in_seconds,
+        )
```
samorojy, line 105: Is it the only one condition we want to cover? 

Initially, I thought we would run certain commands before taking a snapshot, such as bash or something similar. What does the current implementation of `wait_for_pods_ready` wait for? 

In other words, what do we expect to be in this snapshot?

samorojy: It is not about Qodana/Pycharm since they come as a separate plugin.

samorojy (approved):

integration-tests/database/conftest.py
vpoliakov-pixel: true, removed the comment

integration-tests/docker_tests/test-project/.idea/misc.xml
vpoliakov-pixel: just a simulation of a test project with some folders, removed because it belongs to integration tests folder and somehow got into root

orchestrator/src/idegym/orchestrator/snapshot_pipeline.py
vpoliakov-pixel: yes, this branch was created before changing to the generated server name, fixed now

orchestrator/src/idegym/orchestrator/snapshot_pipeline.py
vpoliakov-pixel: Currently it just waits for the healthy state of the pod, meaning that everything is started inside.

It expects that if there is an IDE which is opening a project then the pod would not be healthy until it starts and opens successfully.
But it can be changed later for different checks / instructions.

Initial commands could be put as parts of docker commands or smth, manual triggering of bash commands for "pipeline" does not look appropriate (but it works fine for manual snapshot creation though).

orchestrator/src/idegym/orchestrator/snapshot_pipeline.py
vpoliakov-pixel: extracted into property

PR #129, opened 2026-05-27
[mcp] add mcp session hash, sticky sessions and reduce amount of workers

vpoliakov-pixel:


orchestrator/src/idegym/orchestrator/hydra_configs/orchestrator/base.yaml
dabico: Should we also set the default number of orchestrator workers to match?

dabico (changes requested):
The code changes are okay, what caught my eye were these added annotations. The problems I see with the solution are two-fold:

1. MCP itself becomes a function of the ingress controller. The two are now tightly coupled. Though I wonder to what extent the problems themselves can be solved purely on IdeGYM itself.
2. The chart is written in a generic way so that a client can specify any type of ingress controller they want to use. They may not necessarily want to use NGINX. Furthermore, given its [retirement](https://kubernetes.io/blog/2025/11/11/ingress-nginx-retirement), I'd propose we not exclusively focus on it.

charts/idegym/values.yaml
```diff
@@ -17,7 +17,25 @@ deployment:
...
     tracing: { }
   replicas: 4
-  uvicornWorkers: 4
+  # MCP session state is stored in-memory per uvicorn worker process. As long as
+  # `mcp.statelessHttp` is true every request is independent, so this can be
+  # raised freely. If `mcp.statelessHttp` is set to false, this must stay at 1
+  # or all session-mode requests need to be pinned to one worker.
+  uvicornWorkers: 1
+  mcp:
+    # Force the FastMCP HTTP app to reply with application/json instead of
```
vfedorishchev: Not a comment to block the PR, just a question.
Are you sure this config field forces server to reply with JSON instead of streamable HTTP?
I was under the impression, that this config might be allowing the client to negotiate with the server what response format they support. Therefore, if the server supports both, but client only supports JSON - it will receive the response as JSON. If the client sent the Accepted header of streamable HTTP - the server responds with that.
I didn't personally tested, but I'm curious :)

vpoliakov-pixel: removed this flag completely, with statelessHttp: true it works fine, I guess it was the same issue with sessions but wrapped in a different error

vpoliakov-pixel:
nginx related things are in the comments now; the user is free to choose his own implementation if needed

dabico (approved):

PR #130, opened 2026-05-27
[JBRes-9477] add documentation for pod-snapshot, update Chart

samorojy:
Resolves: JBRes-9477

### DIFF
--- a/charts/idegym/values.yaml
+++ b/charts/idegym/values.yaml
@@ -73,21 +73,23 @@ podSnapshot:
   enabled: false
   serviceAccount:
     create: true
-    name: ""
+    name: "idegym"
     annotations: { }
-  retention:
+  # The keys below mirror the GKE PodSnapshotPolicy / PodSnapshotStorageConfig CRD spec verbatim;
+  # see https://docs.cloud.google.com/kubernetes-engine/docs/concepts/pod-snapshots for the full field reference.
+  snapshotStorageConfig:
+    gcs:
+      bucket: my-idegym-snapshots
+      path: snapshots
+      tokenSource: podKSA
+  retentionConfig:
     lastAccessTimeout: 7d
   snapshotGroupingRules:
-    labels:
-      - idegym.jetbrains.com/snapshot-id
-    maxSnapshotsPerGroup: 5
-  storage:
-    bucket: idegym
-    path: snapshots
-    tokenSource: podKSA
-  trigger:
-    type: manual
-    postCheckpoint: resume
+    groupByLabelValue:
+      labels:
+        - idegym.jetbrains.com/snapshot-id
+      groupRetentionPolicy:
+        maxSnapshotCountPerGroup: 5
 
 postgresql:
   enabled: true
--- a/charts/idegym/values.yaml
+++ b/charts/idegym/values.yaml
@@ -86,8 +86,6 @@ podSnapshot:
     lastAccessTimeout: 7d
   snapshotGroupingRules:
     groupByLabelValue:
-      labels:
-        - idegym.jetbrains.com/snapshot-id
       groupRetentionPolicy:
         maxSnapshotCountPerGroup: 5
 
--- a/charts/idegym/values.yaml
+++ b/charts/idegym/values.yaml
@@ -73,15 +73,15 @@ podSnapshot:
   enabled: false
   serviceAccount:
     create: true
-    name: "idegym"
+    name: ""
     annotations: { }
   # The keys below mirror the GKE PodSnapshotPolicy / PodSnapshotStorageConfig CRD spec verbatim;
   # see https://docs.cloud.google.com/kubernetes-engine/docs/concepts/pod-snapshots for the full field reference.
   snapshotStorageConfig:
     gcs:
-      bucket: my-idegym-snapshots
-      path: snapshots
-      tokenSource: podKSA
+      bucket: ""
+      path: ""
+      tokenSource: ""
   retentionConfig:
     lastAccessTimeout: 7d
   snapshotGroupingRules:
--- a/documentation/remote_deployment.md
+++ b/documentation/remote_deployment.md
@@ -523,3 +523,5 @@ response = httpx.get("https://idegym.yourdomain.com/health", headers=headers)
     - [ ] WIF configured for KSA
     - [ ] Bucket IAM configured
     - [ ] Enable feature gate with `podSnapshot.enabled=true`
+    - [ ] Mirror the bucket and KSA names from the GCP setup in podSnapshot values
+
--- a/documentation/remote_deployment.md
+++ b/documentation/remote_deployment.md
@@ -512,4 +512,4 @@ response = httpx.get("https://idegym.yourdomain.com/health", headers=headers)
 - [ ] `deployment.resources` populated with appropriate requests/limits for your workload
 - [ ] Backup strategy for PostgreSQL persistent volume
 - [ ] gVisor runtime class available on nodes if using sandboxed containers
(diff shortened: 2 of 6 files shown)


charts/idegym/values.yaml
dabico, line 1: Even though I've stated otherwise in the past, I'm a bit on the fence now when it comes to passing CRD spec verbatim as opposed to introducing our own configuration schema. The former gives it more flexibility, but you lose some things as well, like checks and navigation in the editor.

charts/idegym/values.yaml
```diff
@@ -73,21 +73,23 @@ podSnapshot:
   enabled: false
   serviceAccount:
     create: true
-    name: ""
+    name: "idegym"
```
dabico: I wouldn't hard-code this since leaving it empty defaults to a value generated by a helper. This is usually the full name of the release, and `-snapshot` suffix. We use the same generate-name-if-undefined logic for other `ServiceAccount` resources.

charts/idegym/values.yaml
```diff
@@ -73,21 +73,23 @@ podSnapshot:
...
     annotations: { }
-  retention:
+  # The keys below mirror the GKE PodSnapshotPolicy / PodSnapshotStorageConfig CRD spec verbatim;
+  # see https://docs.cloud.google.com/kubernetes-engine/docs/concepts/pod-snapshots for the full field reference.
+  snapshotStorageConfig:
+    gcs:
+      bucket: my-idegym-snapshots
+      path: snapshots
+      tokenSource: podKSA
```
dabico: Should we enforce these defaults? As far as I understood, your implementation should work with other modes like `federatedP4SA`. What I would propose we do instead is not specify anything, leave it up to the user to decide, and add `required` checks to ensure they have done so.

samorojy: The idea was to simplify the set-up process; I don't know how to make it simple and flexible at the same time.

samorojy: Maybe it is not that hard to understand which token source to use. So, agree, will fix.

charts/idegym/values.yaml
samorojy: Agree...
At the moment I do like the flexibility of this approach; it was easy to update the schema without updating everything else

charts/idegym/values.yaml
```diff
@@ -75,19 +75,21 @@ podSnapshot:
...
-  storage:
-    bucket: idegym
-    path: snapshots
-    tokenSource: podKSA
-  trigger:
-    type: manual
-    postCheckpoint: resume
+    groupByLabelValue:
+      labels:
+        - idegym.jetbrains.com/snapshot-id
```
dabico: I just realized that the labels for grouping rules are hard-coded in the manifests of spawned sandboxes. Should we allow this to be configurable? I can see that triggers are also set in stone, but I guess that's fine because of how we work with snapshots.

documentation/remote_deployment.md
```diff
@@ -312,6 +312,74 @@ receive a preferred node affinity and toleration matching the configured key.
...
+
+### Enable the feature
+
+Set `podSnapshot.enabled=true` and point the chart at your bucket. The shape under `podSnapshot.snapshotStorageConfig`,
+`podSnapshot.retentionConfig`, and `podSnapshot.snapshotGroupingRules` mirrors the GKE
+[`PodSnapshotPolicy`](https://docs.cloud.google.com/kubernetes-engine/docs/concepts/pod-snapshots) /
+`PodSnapshotStorageConfig` CRD spec verbatim, so all GKE-documented fields work without translation.
+
+Minimal example — overlay only what differs from the chart defaults in
+[`charts/idegym/values.yaml`](../charts/idegym/values.yaml):
```
dabico, line 347: Think this line needs to be updated to reflect the current state of the values. Also, should the minimal example include WIF annotations?

documentation/remote_deployment.md
```diff
@@ -312,6 +312,74 @@ receive a preferred node affinity and toleration matching the configured key.
...
+```
+
+The chart then renders the `PodSnapshotPolicy`, the `PodSnapshotStorageConfig`, and a Kubernetes `ServiceAccount`
+for snapshot-enabled pods.
+
+It also wires `IDEGYM_POD_SNAPSHOT_ENABLED=True` and `IDEGYM_POD_SNAPSHOT_SERVICE_ACCOUNT_NAME=<KSA>` on the
+orchestrator deployment — the first gates the snapshot endpoints, the second is the KSA the orchestrator attaches
+to every sandbox pod it creates.
+
+> **Notes.** For a snapshot to actually happen and be restorable: the target pod must not run on a GCP E2 node,
```
dabico: To make this stand out, I suggest using admonitions like we do in the rest of the guide:

```suggestion
> [!NOTE]
> For a snapshot to actually happen and be restorable: the target pod must not run on a GCP E2 node,
```

dabico: Another thing about this section: I'd link straight to the [limitations](https://docs.cloud.google.com/kubernetes-engine/docs/concepts/pod-snapshots#limitations) page, since we then don't have to bother with updating it on our end. I guess we should only mention the most impactful limitations here, but you've already done so.

documentation/remote_deployment.md
```diff
@@ -444,3 +512,4 @@ response = httpx.get("https://idegym.yourdomain.com/health", headers=headers)
 - [ ] `deployment.resources` populated with appropriate requests/limits for your workload
 - [ ] Backup strategy for PostgreSQL persistent volume
 - [ ] gVisor runtime class available on nodes if using sandboxed containers
+- [ ] If using pod snapshots: GKE pod-snapshot feature enabled on the cluster, GCS bucket provisioned, snapshot KSA granted the required IAM on the bucket, `podSnapshot.enabled=true` and `snapshotStorageConfig.gcs.bucket` set
```
dabico: I would re-formulate this as a nested checklist, since it's a lot of different steps:

```suggestion
- If using pod snapshots:
    - [ ] GKE pod-snapshot feature enabled on the cluster
    - [ ] GCS bucket provisioned
    - [ ] WIF configured for KSA 
    - [ ] Bucket IAM configured
    - [ ] Mirror configurations for account annotations and bucket in `podSnapshot`
    - [ ] Enable feature gate with `podSnapshot.enabled=true`
```

charts/idegym/values.yaml
samorojy: Good point, but I'd omit that much flexibility for now. 

In any case, for people it would be easy to modify values.yaml and pass desired labels to the `with_server` method.

documentation/remote_deployment.md
samorojy: > WIF annotations

What for? Which annotations do you want to put there.

annotations for GSA sound like an outdated WIf

documentation/remote_deployment.md
samorojy: - [ ] Mirror configurations for account annotations and bucket in `podSnapshot` ?

documentation/remote_deployment.md
dabico: Ah, I re-read their guides and they show IAM binding and not WIF with annotated KSA. Please disregard my comment, it doesn't apply here.

documentation/remote_deployment.md
dabico: It's enough to just mention that the name of the created bucket should be set in values.

samorojy: A. Got it! Thanks.

charts/idegym/values.yaml
dabico: What I meant was: maybe we should inline this label setting into the template directly? At least until it's configurable again. Because right now, if consumers are to override this setting, they run the risk of breaking snapshotting.

dabico (approved):

PR #132, opened 2026-05-29
[JBRes-9587] Pre-release fixes 

samorojy:
Resolves: JBRes-9587

### DIFF
--- a/orchestrator/src/idegym/orchestrator/pod_snapshot.py
+++ b/orchestrator/src/idegym/orchestrator/pod_snapshot.py
@@ -62,15 +62,25 @@ class PodSnapshotService:
         )
         return pod_name, pod_uid, node_name
 
-    def validate_node_eligible_for_snapshot(self, node_name: Optional[str]) -> None:
-        # Heuristic: GKE auto-provisioned node names embed the instance family (e.g.
-        # "gke-...-nap-e2-standard-8-..."). Manually-named pools won't match and will pass through.
-        if node_name and "-e2-" in node_name:
+    async def validate_node_eligible_for_snapshot(self, node_name: Optional[str]) -> None:
+        if not node_name:
+            logger.debug("No node name available; skipping E2 instance type check")
+            return
+
+        async with async_kube_api() as (_, _, core, _, _):
+            node = await core.read_node(name=node_name)
+
+        labels = node.metadata.labels or {}
+        instance_type = labels.get("node.kubernetes.io/instance-type") or labels.get(
+            "beta.kubernetes.io/instance-type"
+        )
+
+        if instance_type and instance_type.startswith("e2-"):
             raise HTTPException(
                 status_code=status.HTTP_400_BAD_REQUEST,
                 detail=(
-                    f"Server is not eligible for snapshot: node '{node_name}' appears to use a GCP E2 "
-                    f"instance type which does not support pod snapshots"
+                    f"Server is not eligible for snapshot: node '{node_name}' uses GCP E2 instance type "
+                    f"'{instance_type}' which does not support pod snapshots"
                 ),
             )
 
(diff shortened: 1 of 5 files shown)


orchestrator/src/idegym/orchestrator/pod_snapshot.py
```diff
@@ -68,23 +62,15 @@ async def get_pod_for_server(self, server_name: str) -> tuple[str, str, str]:
...
-            return
-
-        async with async_kube_api() as (_, _, core, _, _):
-            node = await core.read_node(name=node_name)
-
-        labels = node.metadata.labels or {}
-        instance_type = next((labels[label] for label in _NODE_INSTANCE_TYPE_LABELS if label in labels), None)
-
-        if instance_type and instance_type.startswith("e2-"):
+    def validate_node_eligible_for_snapshot(self, node_name: Optional[str]) -> None:
```
dabico: Since this is a GCP-specific feature, I'd recommend checking GCP-specific node labels like `cloud.google.com/machine-family`. The current heuristic does not cover custom-named node pools.

samorojy: @vpoliakov-pixel

samorojy: As you see, the previous implementation was trying to read node and its labels but it runs into 403 error because deployer SA deployer cannot get resource on the cluster scope. 

We can definitely proceed with ClusterRole + ClusterRoleBinding, but that is not what we want. It will increase the complexity of deployment without huge benefit. 

In the end, that is not something we want to check by ourselves and should be up to the user to understand snapshot limitations and nature. So, the simple heuristic check is left as a handy cheap check.

vpoliakov-pixel (approved):

orchestrator/src/idegym/orchestrator/pod_snapshot.py
dabico: > We can definitely proceed with ClusterRole + ClusterRoleBinding, but that is not what we want. It will increase the complexity of deployment without huge benefit.

How come? I don't see why we can't give `get` permission on `node` for the orchestrator. It won't be able to list all nodes on the cluster. It would only be able to get metadata for nodes by name (and that name can only be obtained from a pod which the orchestrator already sees).

I don't see the benefit in this in its current state.

samorojy: Agree :+1: 
I mean we can give "get" permission, and it will be easy in our setup to empower deployer SA. 

But for the open source version we will require 1 more step which might not be easy for everyone, at the same time snapshot enablement is not easy either. 

So the question is: do we want to handle it manually at all? Especially if Google updates their [limitations](https://docs.cloud.google.com/kubernetes-engine/docs/concepts/pod-snapshots#limitations).

samorojy: Resolution: proceed with ClusterRole for deployer :)

orchestrator/src/idegym/orchestrator/pod_snapshot.py
```diff
@@ -140,10 +136,8 @@ async def wait_for_completion(self, trigger_name: str) -> None:
             if condition:
                 triggered = condition.get("status") == "True"
                 if triggered:
-                    internal_name = ((obj.get("status") or {}).get("snapshotCreated") or {}).get("name") or ""
-                    logger.info(
-                        f"PodSnapshotManualTrigger '{trigger_name}' completed, snapshot name: '{internal_name}'"
-                    )
+                    logger.info(f"PodSnapshotManualTrigger '{trigger_name}' completed")
+                    time.sleep(3)   # Delay to ensure image got uploaded to the GCS
```
dabico: Please use `asyncio.sleep` for non-blocking waiting.

dabico (changes requested):

orchestrator/src/idegym/orchestrator/pod_snapshot.py
samorojy: Thanks!!

dabico (approved):

PR #141, opened 2026-06-02
[JBRes-9093] extract cleanup watcher into a standalone service

vpoliakov-pixel:
The watcher cleanup loop (inactive servers/clients, stale request records, orphaned Kaniko jobs) ran inside the orchestrator process, which was causing troubles when the orchestrator has multiple workers. Extract it into a dedicated single-replica service so the orchestrator can scale freely.

- New `idegym-watcher` workspace package: reuses the orchestrator DB layer, nodes_holder and k8s client; minimal FastAPI app exposing /health with the cleanup loop in its lifespan. Connects to the DB without running migrations (orchestrator owns the schema) via the extracted `connect_db_engine`.
- Remove the watcher startup from the orchestrator; move watcher.py to the new package as cleanup.py (advisory lock kept as a rolling-update safety net).
- Helm: dedicated watcher Deployment with distinct labels, ServiceAccount and RoleBinding to the existing Role; shared POSTGRES_* env extracted into the `idegym.databaseEnv` helper (orchestrator render unchanged).
- Build script + CI publish job for the watcher image; exclude it from PyPI.
- Tests: DB-level + Docker-image integration tests and a functional e2e test that asserts the standalone watcher reaps a stale server in-cluster.

Resolves: JBRes-9093


vfedorishchev (approved):

charts/idegym/values.yaml
dabico: Since the same database is accessed by two services, would it make more sense to place this configuration at the top level? How it's defined now seems like it's exclusive to the orchestrator, when in reality it's both it and the watcher.

backend-utils/src/idegym/backend/utils/logging.py
dabico, line 1: What prompted this change now? I see what's in the comment, but the described issue to me reads like it pre-dates this watcher refactoring. In any case, I think it's something that should be addressed properly.

charts/idegym/templates/watcher-deployment.yaml
```diff
@@ -0,0 +1,140 @@
...
+    matchLabels:
+      {{- include "idegym.watcher.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "idegym.watcher.selectorLabels" . | nindent 8 }}
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: "/metrics"
+        prometheus.io/port: "{{ .Values.service.containerPort }}"
```
dabico, line 20: There are currently no metrics exposed by the watcher.

watcher/supervisord.conf
dabico, line 1: Does watcher really need `supervisord`?

charts/idegym/values.yaml
vpoliakov-pixel: Extracted to the top level

charts/idegym/templates/watcher-deployment.yaml
vpoliakov-pixel: Added metrics

backend-utils/src/idegym/backend/utils/logging.py
vpoliakov-pixel: Yes, this whole thing appeared after starting playing around uvicorn workers and the differences between having 1 or multiple workers in terms of processes, log formatting and available fields.
No cleaner solution was found, especially because having `configure_logging` method idempotent is much better than adding some if-s here and there. We use this method everywhere - in orchestrator, IdeGYM server and watcher.
We can try to refactor logging in general separately.

watcher/supervisord.conf
vpoliakov-pixel: No, there is no specific need for supervisord for watcher, it was introduced just to have everything run similarly.
We can also refactor it later as a whole, in a separate MR because it is not actually needed e.g. for orchestrator too

PR #159, opened 2026-06-11
[JBRes-9632] restore a server from a specific pod snapshot via snapshot-tag

vpoliakov-pixel:
## What & why
There is a need to restore a server from a *specific* GKE pod snapshot, not only the latest. Today restore is driven by the `idegym.jetbrains.com/snapshot-id` label and GKE always picks the latest snapshot in the group.

## Changes
- New `snapshot_tag` on `StartServerRequest` (requires `snapshot_id`). When set, `deploy_server` writes the pod annotation `podsnapshot.gke.io/ps-name: <tag>` (GKE restore-specific-snapshot); `snapshot_id` alone keeps latest-in-group.
- Capture the GKE-generated PodSnapshot name from the `PodSnapshotManualTrigger` status, persist it in new nullable column `snapshots.pod_snapshot_name` (Alembic 003), and return it as `snapshot_tag` on create/prepare/job-status/exists responses so a tag round-trips through the API.
- Threaded through the full client -> operations -> request -> router -> deploy_server chain.


api/src/idegym/api/orchestrator/servers.py
```diff
@@ -91,6 +91,22 @@ class StartServerRequest(BaseModel):
             "Leave it empty to start a new server."
         ),
     )
+    snapshot_tag: Optional[str] = Field(
+        default=None,
+        description=(
+            "GKE ONLY: Restore a specific snapshot instead of the latest one. "
+            "The value is the GKE PodSnapshot resource name (see the snapshot_tag returned "
+            "when the snapshot was created), passed through as the 'podsnapshot.gke.io/ps-name' "
+            "pod annotation. Must be specified together with snapshot_id; leave empty to restore "
```
vfedorishchev: If I get it right snapshot_tag is only valid if the snapshot_id is present. What do you think about aggregating those two fields into one like:

class Snapshot(BaseModel):
snapshot_id: str <-- this will be required at all times if the Snapshot is requested
snapshot_tag: Optional[str] <-- stays optional

and then in here

class StartServerRequest(BaseModel):
//....

snapshot: Optional[Snapshot] = Field()

this way you won't need the validator underneath

api/src/idegym/api/orchestrator/snapshots.py
```diff
@@ -56,6 +56,14 @@ class CreateSnapshotResponse(BaseModel):
         default=None,
         description="ID of the taken snapshot. By implementation, equals the snapshot_id / name of the snapshotted pod.",
     )
+    snapshot_tag: Optional[str] = Field(
```
vfedorishchev, line 111: and if you apply the comment above - we can probably re-use the class across the fileds

orchestrator/src/idegym/orchestrator/pod_snapshot.py
```diff
@@ -185,3 +189,21 @@ def _triggered_condition(obj: dict[str, Any]) -> dict[str, Any] | None:
             if cond.get("type") == "Triggered":
                 return cond
         return None
+
+    @staticmethod
+    def _created_snapshot_name(obj: dict[str, Any]) -> Optional[str]:
```
vfedorishchev: I first thought it's our design to use dicts without defined fields and types, but when I saw it's k8s lib that does it, I praised our coding quality a bit more than usual.

orchestrator/src/idegym/orchestrator/pod_snapshot.py
```diff
@@ -185,3 +189,21 @@ def _triggered_condition(obj: dict[str, Any]) -> dict[str, Any] | None:
...
+        """Extract the auto-generated PodSnapshot resource name from a completed trigger status.
+
+        GKE's docs disagree on the shape, so handle all observed forms: status.snapshotCreated
+        as a plain string (CRD reference), as a nested object with a `name` field (how-to guides),
+        or a flat status.snapshotCreatedName. Returns None when the field is absent.
+        """
+        status = obj.get("status") or {}
+        snapshot_created = status.get("snapshotCreated")
```
vfedorishchev: I think you are not handling the None case correctly?

api/src/idegym/api/orchestrator/servers.py
vpoliakov-pixel: fixed

orchestrator/src/idegym/orchestrator/pod_snapshot.py
vpoliakov-pixel: fixed

api/src/idegym/api/orchestrator/snapshots.py
```diff
@@ -43,6 +43,51 @@ class PodSnapshotManualTrigger(BaseModel):
...
+
+class PodSnapshotManualTriggerStatus(BaseModel):
+    """Typed view over the otherwise-untyped PodSnapshotManualTrigger.status dict the k8s API returns."""
+
+    model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True, extra="ignore")
+
+    conditions: list[PodSnapshotTriggerCondition] = Field(default_factory=list)
+    # GKE's docs disagree on the shape: a nested object carrying `name`, or the plain name string.
+    snapshot_created: Optional[Union[PodSnapshotCreated, str]] = None
+    snapshot_created_name: Optional[str] = None
```
samorojy, line 76: https://github.com/jetbrains-research/idegym/pull/118#issue-4496026997 

Did you take this schema into consideration?

samorojy (approved):
LGTM, but let's check on us-central1

api/src/idegym/api/orchestrator/snapshots.py
samorojy: Fixed.

samorojy:
Verified on us-central1. 

At the moment it is deployed there, feel free to test it as well.

samorojy:
@vpoliakov-pixel 
Check the changes. I think it is ready.

vpoliakov-pixel:
LGTM, let's combine both migrations into one because it is one MR and one feature so they should come simultaneously, and we are ready to merge

samorojy:
Let’s wait for for tests

PR #162, opened 2026-06-12
[JBRes-9066] Handle pod crash events: detect crashes, surface reason, restart budget

vpoliakov-pixel:
## Problem

Server pods run as Kubernetes **Deployments** (`replicas=1`), whose pod templates are forced to `restartPolicy: Always`. When a container dies — OOM, non-zero exit, or the gvisor **out-of-storage** kernel failure — the kubelet restarts it **forever**. The DB row stays `ALIVE`, the crash reason is never recorded, and nothing breaks the loop. One bad model command can stall training for hours with no error surfaced to the user.

Nothing watched pod lifecycle before this: the watcher service (extracted in #141) only did timeout-based cleanup.

## Solution

**1. Crash detection in the watcher** — new `watcher/src/idegym/watcher/crash_detector.py`
- `evaluate_pod_crash(pod, max_restarts)` (pure): a crash is `restart_count > budget`, or pod-level `phase=Failed` / `reason=Evicted` (the disk-pressure / out-of-storage case). Builds a human-readable reason exposing `OOMKilled` / exit codes / `CrashLoopBackOff`.
- `detect_crashed_servers(db)`: fetches `ALIVE`+`REUSED` servers and lists pods **once per namespace** (`app.kubernetes.io/part-of=idegym`, indexed by the `app` label = `generated_name`) — never per-server, **no Events API**. Runs inside the existing cleanup loop under its Postgres advisory lock, gated on `WatcherConfig.crash_detection_enabled` (default on). Detection latency ≈ one `cleanup_interval` (60s).

**2. Surface the reason + stop the loop**
- On crash: **delete the Deployment first** (`clean_up_server`), then mark `CRASHED` on success / `DELETION_FAILED` on teardown failure — mirroring `cleanup_servers`. Both record the reason in the new `IdeGYMServer.details` column (migration `003`).
- `validate_server` appends `details` to its `410 GONE` response, so the client sees **why** on the next forward instead of an opaque error. Terminal status frees the resource quota.

**3. Per-server restart budget**
- `StartServerRequest.max_restarts`, default **0** (fail on first crash), plumbed client → api → orchestrator → DB (new `servers.max_restarts` column), following the established `snapshot_id` chain. Available on `start_server` and `with_server`. Raise it to let transient crashes self-heal.

## Tests
- `unit-tests/test_crash_detector.py` — `evaluate_pod_crash` table-driven over fabricated pods (OOM, eviction, CrashLoopBackOff, within/over budget, terminating, missing status) + `detect_crashed_servers` with mocked k8s/db (one-list-per-namespace, teardown-failure → `DELETION_FAILED`).
- `integration-tests/database/test_watcher_crash.py` — testcontainers Postgres + mocked k8s: OOM → `CRASHED`+details+torn down; within budget → stays `ALIVE`; eviction → `CRASHED`; teardown failure → `DELETION_FAILED`+details.

**Verification:** 522 unit + 86 DB integration tests pass; `ruff` clean; migration chain `001→002→003` (single head) verified end-to-end on a real Postgres via the project's `MigrationManager`.

## Notes
- No bare-Pod / `restartPolicy: Never` mode — Deployments forbid it and it would break the PDB / snapshot-restore / deploy flow. The restart **budget** delivers equivalent control within the Deployment model.
- With `max_restarts=0`, a pod that crashed once and self-recovered is still marked `CRASHED` (intended fail-fast default — the crash is surfaced rather than hidden).


watcher/src/idegym/watcher/crash_detector.py
```diff
@@ -0,0 +1,155 @@
...
+    status = pod.status
+    if status is None:
+        return None
+
+    if status.phase == "Failed" or status.reason == "Evicted":
+        reason = status.reason or "Failed"
+        message = status.message.strip() if status.message else _terminated_description(pod)
+        description = f"Pod {reason.lower()}"
+        return f"{description}: {message}" if message else description
+
```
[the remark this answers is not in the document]
vpoliakov-pixel: The guard is intentional, and Copilot's premise is wrong on the Kubernetes semantics:

  1. Node-pressure evictions (the out-of-storage case this feature targets) do NOT set deletion_timestamp. The kubelet evicts by setting status.phase=Failed / status.reason=Evicted and killing the containers — it leaves the pod object
  in place (that's exactly why Evicted pods notoriously pile up under kubectl get pods). So those pods have deletion_timestamp = None, sail past the guard, and are already caught by the phase == "Failed" or reason == "Evicted" branch.
  The test_evicted_pod_marks_server_crashed integration test exercises precisely this. So "common for evictions [to be marked for deletion]" is false for the case we care about.

  2. The only pods that carry deletion_timestamp are ones being deliberately torn down — our own clean_up_server, a client stop/finish, or the /restart endpoint (which keeps the server ALIVE while cycling its pods). If I reorder so
  Failed/Evicted is evaluated before the deletion guard, then a pod that exits non-zero while handling SIGTERM during one of those legitimate teardowns would momentarily look Failed, and the watcher would race in and stamp the server
  CRASHED mid-restart/stop. That's a real regression — a false terminal status on a server the user is intentionally restarting.

  So reordering trades a correctly-handled case (evictions, already caught) for a new false-positive on the restart/stop path. The narrow case Copilot imagines — a server whose sole pod is a Failed/Evicted pod that also has a
  `deletion_timestamp* and has no replacement yet — is transient (the ReplicaSet brings up a fresh pod; if the problem persists, that pod crashes too and is caught next cycle) and isn't worth that regression.

watcher/src/idegym/watcher/crash_detector.py
```diff
@@ -0,0 +1,155 @@
...
+from idegym.backend.utils.utils import log_exceptions
+from idegym.orchestrator.database.database import (
+    get_idegym_servers_by_status,
+    update_idegym_server_heartbeat,
+)
+from idegym.utils.logging import get_logger
+from sqlalchemy.ext.asyncio import AsyncSession
+
+if TYPE_CHECKING:
+    from kubernetes_asyncio.client import V1Pod
```
samorojy, line 16: Do you think that we are building our project the right way if we do need these hacks to import without cycles?

samorojy (approved):

watcher/src/idegym/watcher/crash_detector.py
vpoliakov-pixel: most likely could be removed completely in any case, it it not related to building the project, more likely about not importing a lot of stuff if not needed

PR #182, opened 2026-06-21
[JBRes-9821] fix security context at helm chart for watcher

samorojy:
Watcher hasn't been able to start without a root. 

Resolves: JBRes-9821

### DIFF
--- a/charts/idegym/values.yaml
+++ b/charts/idegym/values.yaml
@@ -72,7 +72,10 @@ watcher:
       prePing: true
     maxOverflow: ""
   resources: { }
-  securityContext: { }
+  securityContext:
+    runAsNonRoot: true
+    runAsUser: 1000
+    runAsGroup: 1000
   serviceAccount:
     create: true
     name: ""


samorojy:
Should we patch the published helm chart?

charts/idegym/values.yaml
```diff
@@ -72,7 +72,10 @@ watcher:
       prePing: true
     maxOverflow: ""
   resources: { }
-  securityContext: { }
+  securityContext:
+    runAsNonRoot: true
+    runAsUser: 1000
+    runAsGroup: 1000
```
dabico, line 78: How come we need this if we use the dedicated group/user for running the watcher?

https://github.com/JetBrains-Research/idegym/blob/04e63c23d5f374afb6f95174175001f2dc64587d/watcher/Dockerfile#L44

dabico: Furthermore, these settings should not be the enforced default. Most environments don't have strict security policies that prohibit running as the `root` user. This is also why things worked without issue in E2E tests. I would keep this setting as it was before. If your infrastructure requires non-`root`, then it's up to you to override the values.

dabico (changes requested):

charts/idegym/values.yaml
samorojy: Agree that we should not force root for the default chart. 

Regarding the security context:

The image only records the user string "watcher", not 1000. The mapping watcher → 1000 lives inside the image's /etc/passwd file. The kubelet must prove the user is non-root before it starts the container.

To do that, kubelet has to mount image fs and read passwd and parse watcher:x:1000, but it doesn't do this. Resolving a username happens inside the container at startup. 

So, we pass security context to let kubelet know the UID.

PR #230, opened 2026-07-14
[JBRes-9808] Agentless OpenHands tools plugin (REST + MCP + typed client)

vpoliakov-pixel:
Resolves: JBRes-9808

## What

Adds an **agentless OpenHands tools plugin** to IdeGym (`plugins/openhands/`, part of the
`idegym-plugins` distribution). It exposes the agent-independent parts of
[`openhands-tools`](https://github.com/OpenHands/software-agent-sdk) — terminal, file editor, grep,
glob, apply-patch, planning file editor, task tracker, and the Gemini file tools — through **three
equal surfaces** (REST `/api/openhands/*`, MCP `openhands` namespace, and a typed client
`server.openhands`), all backed by **one** runtime and **one** set of stateful terminal sessions.

There is no OpenHands agent, LLM, CLI, or conversation loop. The plugin reuses OpenHands' executors,
tool definitions, action/observation models, MCP schema generation, and terminal sessions — it is
purely the IdeGym integration glue.

## How it works

```mermaid
flowchart TB
    caller(["<b>server.openhands.*</b><br/>typed client"]):::client
    rest(["<b>/api/openhands/*</b><br/>REST"]):::client
    mcp(["<b>/mcp</b><br/>openhands namespace"]):::client

    subgraph container["IdeGym task container"]
        splugin[["server plugin<br/>(typed proxy)"]]:::pod
        gateway[["FastMCP gateway"]]:::pod
        subgraph venv["dedicated venv · 127.0.0.1"]
            svc[["FastAPI /v1 + FastMCP /mcp"]]:::pod
            runtime{{"<b>ToolRuntime</b>"}}:::ctrl
            term("TerminalSessionManager<br/>tmux · subprocess"):::tool
            adapters("OpenHands adapters"):::tool
        end
    end
    oh[("openhands-sdk · openhands-tools")]:::store

    caller --> splugin
    rest --> splugin
    mcp --> gateway
    splugin -->|"loopback HTTP"| svc
    gateway -->|"proxy"| svc
    svc --> runtime
    runtime --> term
    runtime --> adapters
    term --> oh
    adapters --> oh

    classDef client fill:#2563eb,stroke:#1d4ed8,color:#fff;
    classDef ctrl fill:#6b57ff,stroke:#5b4bd2,color:#fff;
    classDef pod fill:#4f46e5,stroke:#4338ca,color:#fff;
    classDef tool fill:#e23b3b,stroke:#c02626,color:#fff;
    classDef store fill:#0891b2,stroke:#0e7490,color:#fff;
```

## Key design decisions

- **Own in-container venv, sourced like the other plugins.** `openhands-sdk` transitively pins
  `opentelemetry-api==1.39.1` (via `lmnr`) while `idegym-backend-utils` needs `>=1.43.0` — mutually
  exclusive. So OpenHands is kept out of the workspace lock and the service runs in a dedicated venv;
  IdeGym only proxies over loopback (this is why it is a separate process). The venv installs the
  plugin runtime from the `idegym-plugins` source the base image already provides (`$IDEGYM_PATH/plugins`) —
  the same distribution the idea/pycharm plugins come from, nothing fetched or copied twice — with
  `package_spec` as the override to install from PyPI once `idegym-plugins` is published there.
  Nothing installs at container startup; the runtime is baked at build time.
- **Two terminal backends, tmux recommended, never silently switched.** Terminals are created via
  OpenHands' `create_terminal_session` (one retained pinned pane/process per handle); the backend is
  fixed per terminal and reported in every result, and a disabled/unavailable backend errors
  explicitly. **tmux is the reliable backend** — OpenHands' subprocess terminal has an unreliable
  interrupt (upstream warns to install tmux), so deployments that interrupt long-running commands
  should use tmux (the image installs it when that backend is enabled; `has_foreground_command` reads
  OpenHands' authoritative `is_running()`). A native retained-PTY subprocess shell is used only when
  OpenHands is absent (dev/CI) so the service still runs and is testable.
- **One runtime, three projections.** REST, MCP, and the client all dispatch to the same
  `ToolRuntime` and the same terminal ids — a terminal created on one surface is usable from another.
- **Nothing faked.** Agent-dependent families (`task`, `workflow`, `tom_consult`) are reported
  `unsupported_requires_agent`, `delegate`/`preset`/`utils` `not_a_callable_tool`, and a missing
  optional dependency `missing_dependency` — never silently omitted or emulated.

## Tests

| Suite | Result |
|---|---|
| unit (`unit-tests/test_openhands_*.py`) | 65 passed — native subprocess terminal contract, catalog audit, REST+MCP + cross-surface |
| compat (`plugins/openhands/run-compat-tests.sh`) | 5 passed + 1 skipped (tmux-gated interrupt-then-usable) against real `openhands-tools==1.36.0` in an isolated venv |
| integration (Docker) | 9 passed — real image builds (subprocess + tmux) |
| e2e (`e2e-tests/test_openhands.py`) | **green** on CI/minikube — 4 tests build + deploy + exercise all three surfaces on the tmux backend: stateful terminals (cwd/env/foreground input/interrupt/reset/isolation) + file/search/patch tools |

## CI

Splits the e2e job into a **balanced parallel matrix** — `idea`, `pycharm`, `openhands`, `kaniko`,
`mcp`, `other` — differing only by the pytest `-m` expression, and disables `fail-fast` on the test
and e2e matrices. Group markers are auto-applied by file name in `conftest.py`; `other` is a catch-all
negation so a new e2e file is never dropped.

> [!IMPORTANT]
> CI check names change. Update branch-protection required checks: remove
> `CI / Run e2e tests (excluding IDE integrations)` and `CI / Run e2e IDE integration tests`; add
> `CI / Run e2e {idea,pycharm,openhands,kaniko,mcp,other} tests`.


plugins/openhands/src/idegym/plugins/openhands/image.py
```diff
@@ -0,0 +1,197 @@
...
+        chown_targets = f"{dirs} {self.venv_dir}" if self.install_openhands else dirs
+        parts += [
+            "",
+            "RUN set -eux; \\",
+            f"    mkdir -p {dirs}; \\",
+            f"    chown -R {user}:{user} {chown_targets}",
+            "",
+            f"COPY {_START_SCRIPT_DEST} /usr/local/bin/start-openhands-service.sh",
+            f"COPY {_SUPERVISOR_DEST} /etc/supervisor/conf.d/openhands.conf",
+            "RUN chmod +x /usr/local/bin/start-openhands-service.sh",
```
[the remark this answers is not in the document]
vpoliakov-pixel: Fixed in a811290. The image plugin now `COPY`s the start script to the bare command name `/usr/local/bin/start-openhands-service` (not `.sh`) and `chmod +x`es it, so the supervisor `command=/usr/local/bin/start-openhands-service` is valid even in images that don't run IdeGYMServer's `/usr/local/bin/*.{py,sh}` rename pass.

integration-tests/docker_tests/test_openhands_plugin.py
```diff
@@ -0,0 +1,155 @@
...
+@pytest.mark.integration
+def test_service_assets_are_present():
+    check = (
+        "test -f /etc/supervisor/conf.d/openhands.conf && "
+        "test -f /usr/local/bin/start-openhands-service.sh && "
+        "test -d /var/lib/idegym-openhands/state && "
+        "test -d /var/lib/idegym-openhands/artifacts && "
+        "test -f /etc/idegym/mcp-upstreams.d/openhands.json && "
+        "echo ALLOK"
+    )
```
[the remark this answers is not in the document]
vpoliakov-pixel: Fixed in a811290. The script is now installed to the bare name `/usr/local/bin/start-openhands-service`, and this test asserts `test -x /usr/local/bin/start-openhands-service`.

plugins/openhands/src/idegym/plugins/openhands/runtime/service.py
```diff
@@ -0,0 +1,436 @@
...
+
+    def health(self) -> HealthResponse:
+        workspace_ok = Path(self.config.workspace_root).is_dir()
+        backend_ok = self.terminals.default_backend_ready()
+        checks = {
+            "workspace": workspace_ok,
+            "default_backend": backend_ok,
+            "catalog": len(self.catalog.entries()) > 0,
+            "artifacts_dir": Path(self.config.output_dir).is_dir() or True,
+        }
```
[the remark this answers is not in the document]
vpoliakov-pixel: Fixed in a811290. The check is now `Path(self.config.output_dir).is_dir()` (the output dir is created in `prepare()`), so `/v1/health` and `/readyz` fail meaningfully when the artifact directory is missing.

plugins/openhands/src/idegym/plugins/openhands/runtime/terminal/backends/openhands.py
```diff
@@ -0,0 +1,118 @@
...
+from idegym.plugins.openhands.api.models import TerminalBackend
+from idegym.plugins.openhands.runtime import compat
+from idegym.plugins.openhands.runtime.terminal.backend import BackendExec, BackendHealth, TerminalBackendSession
+
+
+class OpenHandsTerminalSession(TerminalBackendSession):
+    """Adapts a retained OpenHands terminal session to the backend contract."""
+
+    capture_supported = True
+
```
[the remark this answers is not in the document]
vpoliakov-pixel: Fixed in a811290. The OpenHands-backed terminal session now sets `capture_supported = False` (OpenHands terminal sessions expose no pane/buffer capture; callers should read output from execute/poll). The native subprocess backend keeps `capture_supported = True`, and the descriptor reflects the actual backend.

plugins/openhands/src/idegym/plugins/openhands/api/models.py
```diff
@@ -0,0 +1,355 @@
...
+class TerminalCreateRequest(_Strict):
+    """Body for ``POST /api/openhands/terminals``."""
+
+    backend: Optional[TerminalBackend] = None
+    name: Optional[str] = None
+    cwd: Optional[str] = None
+    env: dict[str, str] = Field(default_factory=dict)
+    no_change_timeout: Optional[float] = None
+    cols: Optional[int] = None
+    rows: Optional[int] = None
```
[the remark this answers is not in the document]
vpoliakov-pixel: Fixed in a811290. `cols`/`rows` are now passed through to the subprocess backend's PTY, and `no_change_timeout` becomes the handle's default no-change timeout for execute/input (and the OpenHands session's `no_change_timeout_seconds`). All three are stored on the handle and re-applied on reset.

plugins/openhands/src/idegym/plugins/openhands/api/errors.py
```diff
@@ -0,0 +1,87 @@
...
+    TOOL_REQUIRES_AGENT = "tool_requires_agent"
+    QUOTA_EXCEEDED = "quota_exceeded"
+    SERVICE_UNAVAILABLE = "service_unavailable"
+    DEADLINE_EXCEEDED = "deadline_exceeded"
+    INTERNAL_ERROR = "internal_error"
+
+
+# Mapping from error code to suggested HTTP status. Command exit codes never
+# appear here: a nonzero shell command is a 200 response with ``is_error`` metadata.
+_HTTP_STATUS: dict[ErrorCode, int] = {
```
vfedorishchev: I think this better be a part of the enum itself. It supports methods like in the example from the official docs https://docs.python.org/3/library/enum.html#enum.Enum.__dir__

plugins/openhands/src/idegym/plugins/openhands/api/errors.py
```diff
@@ -0,0 +1,87 @@
...
+    ErrorCode.INTERNAL_ERROR: 500,
+}
+
+
+def http_status_for(code: ErrorCode) -> int:
+    """Return the suggested HTTP status for an error code (defaults to 500)."""
+    return _HTTP_STATUS.get(code, 500)
+
+
+class ErrorResponse(BaseModel):
```
vfedorishchev: is there any value in this (data) class?

I see it's only being created when the service error is thrown and then immediately used for json_dump call.
With such an approach, you can have a to_json methon in the ServiceError itself and return a readily serealized object back. Unless you foresee a scenario where the ErrorResponse would be useful all by itself and we should keep it.

plugins/openhands/src/idegym/plugins/openhands/runtime/adapters/openhands.py
```diff
@@ -0,0 +1,51 @@
...
+
+from typing import Any, Optional
+
+from idegym.plugins.openhands.api.errors import ErrorCode, ServiceError
+from idegym.plugins.openhands.api.models import ContentBlock, ContentType
+from idegym.plugins.openhands.runtime import compat
+from idegym.plugins.openhands.runtime.adapters.base import AdapterRun
+
+
+class OpenHandsToolAdapter:
```
vfedorishchev: is it supposed to be inherited from `ToolAdapter`?

plugins/openhands/src/idegym/plugins/openhands/runtime/adapters/openhands.py
```diff
@@ -0,0 +1,51 @@
...
+        mcp = compat.tool_mcp_schema(tool)
+        self.name: str = mcp.get("name", getattr(tool, "name", "unknown"))
+        self.description: str = mcp.get("description", getattr(tool, "description", "") or "")
+        self.input_schema: dict[str, Any] = mcp.get("inputSchema", {})
+        self.output_schema: Optional[dict[str, Any]] = mcp.get("outputSchema")
+        self.annotations: dict[str, bool] = compat.tool_annotations(tool)
+
+    async def run(self, arguments: dict[str, Any]) -> AdapterRun:
+        try:
+            action = self._tool.action_from_arguments(arguments)
```
vfedorishchev, line 30: I would really like to see some sort of interface that would ensure the compatibility with OpenHands in compile time. Not just for this particular dynamic call, but for the rest of OpenHands invocations via what currently is an `Any` type

plugins/openhands/src/idegym/plugins/openhands/runtime/service.py
```diff
@@ -0,0 +1,436 @@
...
+                ErrorCode.TOOL_DISABLED,
+                f"Tool {entry.name} is not currently callable (openhands-tools not installed)",
+            )
+        # Path policy for file tools: normalize + reject workspace escapes.
+        if entry.lock_scope == LockScope.PATH:
+            self._extract_path(arguments)
+        call_id = _new_call_id()
+        started = _now()
+        async with self.scheduler.acquire(self._lock_keys(entry, arguments)):
+            run = await adapter.run(arguments)
```
vfedorishchev: I think we might face an issue here and in other places that invoke the adapter. The underlying code of **[OpenHands](https://github.com/OpenHands/software-agent-sdk/blob/51c102b9c0348bbdd4e6a84b1ac4199e0d77f827/openhands-sdk/openhands/sdk/tool/tool.py#L392)** starts an asyncio thread. Which (if I get it right) doesn't have a timeout, shared lifespan or anything like this.
Meaning, if the request is cancelled - the thread with that asyncio task will keep on running, modifying the data behind the scenes (potentially).
The lock mechanism won't work since it's going to be released by the ContextManager.

Probably safeguarding from that here and in other places where it could happen makes sense.

vfedorishchev (approved):

plugins/openhands/src/idegym/plugins/openhands/api/errors.py
vpoliakov-pixel: no value, removed

plugins/openhands/src/idegym/plugins/openhands/api/errors.py
vpoliakov-pixel: fixed

plugins/openhands/src/idegym/plugins/openhands/runtime/adapters/openhands.py
vpoliakov-pixel: yes, thank you!

plugins/openhands/src/idegym/plugins/openhands/runtime/adapters/openhands.py
vpoliakov-pixel: not really possible because openhands is not a real dependency for the project, but now it is a bit better

plugins/openhands/src/idegym/plugins/openhands/runtime/service.py
vpoliakov-pixel: fixed

PR #231, opened 2026-07-14
[JBRes-10122] Run IDEA with a virtual display (headless flag)

vpoliakov-pixel:
## JBRes-10122 — Run IDEA with a virtual display

Adds a `headless` flag to the `Idea` image plugin. `Idea(headless=True)` (default) keeps true headless mode; `Idea(headless=False)` runs the IDE under an Xvfb virtual display on `:99` — the same mechanism PyCharm always uses — for workloads that need a real AWT toolkit (Swing UI rendering, plugins that break under headless AWT).

### How it works
The flag is threaded the same way `MCP_STEROID` is: a build-time env baked into the image, read by the start script.

- **`image.py`** — new `headless: bool = True` field, passed into the install template.
- **`Dockerfile.install.j2`** — when `headless=False`: appends `xvfb xdotool` to apt, sets `ENV DISPLAY=":99"`, and **omits** `-Djava.awt.headless=true` from the vmoptions (forcing headless would make the IDE ignore the Xvfb display). Always bakes `ENV IDEA_HEADLESS=true|false`.
- **`start-idea.sh`** — reads `IDEA_HEADLESS` (default `true`): headless → `JAVA_TOOL_OPTIONS` + `-Djava.awt.headless=true`; non-headless → start `Xvfb :99` (same block as `start-pycharm.sh`) and kill it on shutdown.

`inspect.sh` is unaffected — JetBrains batch inspect forces headless itself (as it already does for PyCharm, whose vmoptions never carried the flag).

### Tests
- **Unit** — `test_idea_headless_by_default`, `test_idea_render_virtual_display_installs_xvfb`.
- **E2E** — `e2e-tests/test_idea_virtual_display.py` (marker `ide_integrations`), modelled on `test_pycharm_mcp_steroid.py`: builds `Idea(headless=False, mcp_steroid=True)`, asserts mcp-steroid is ready, verifies the IDE genuinely runs under the virtual display (`pgrep -x Xvfb`, `DISPLAY=:99`, `IDEA_HEADLESS=false`), then uses `steroid_list_windows` to confirm a **rendered GUI window** (impossible headless) and opens java-project. Runs in the `e2e-ide-integrations` CI job.
- The shared mcp-steroid window/screenshot/dialog helpers were extracted into `e2e-tests/utils/mcp_steroid.py`; `test_pycharm_mcp_steroid.py` now imports them (behavior-preserving refactor).

### Commits
1. `[JBRes-10122] Add headless flag to run IDEA under a virtual display` — feature + unit tests
2. `[JBRes-10122] Document IDEA headless/virtual-display flag` — README + website docs
3. `[JBRes-10122] Extract shared mcp-steroid e2e window helpers` — util + PyCharm test refactor
4. `[JBRes-10122] Add IDEA virtual-display e2e test` — e2e test + `ide_integrations` marker doc fix

### Not verified locally
The full e2e needs the minikube IDE harness + an ~800 MB IDEA download; it's validated by collection + correctness-by-construction against the proven PyCharm/Xvfb path. Unit tests, lint, format, template rendering, and shell syntax all pass locally.


e2e-tests/utils/mcp_steroid.py
```diff
@@ -0,0 +1,163 @@
...
+import asyncio
+import base64
+import json
+import os
+from importlib.resources import files
+from typing import Optional
+
+import resources as e2e_resources
+from idegym.utils.logging import get_logger
+from pydantic import BaseModel, ConfigDict
```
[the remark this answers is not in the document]
vpoliakov-pixel: Done in 87e8d63 — imported `Field` and switched `windows` to `Field(default_factory=list)`.

e2e-tests/utils/mcp_steroid.py
```diff
@@ -0,0 +1,163 @@
...
+    modal_dialog_showing: bool = False
+    project_name: Optional[str] = None
+    bounds: Optional[WindowBounds] = None
+    project_path: Optional[str] = None
+    window_id: Optional[str] = None
+    id: Optional[str] = None
+
+
+class McpWindowsResult(BaseModel):
+    windows: list[IdeWindow] = []
```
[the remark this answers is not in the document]
vpoliakov-pixel: Fixed in 87e8d63 (`Field(default_factory=list)`). For the record, pydantic v2 already deep-copies field defaults per instance, so there was no actual cross-instance leak — but the explicit factory is the idiomatic form and removes the doubt.

plugins/idea/scripts/start-idea.sh
```diff
@@ -65,30 +82,49 @@ else
...
+# IDEA supports -Djava.awt.headless=true natively (the default). When
+# IDEA_HEADLESS=false, Xvfb provides a virtual X11 display that satisfies AWT
+# without real hardware — the same way PyCharm always runs.
+XVFB_PID=""
+HEADLESS_ARGS=()
+if [ "${IDEA_HEADLESS}" = "true" ]; then
+    # JAVA_TOOL_OPTIONS is read by the JVM before application startup code, making
+    # it the most reliable way to set java.awt.headless=true.
+    export JAVA_TOOL_OPTIONS="-Djava.awt.headless=true"
+    HEADLESS_ARGS=("-Djava.awt.headless=true")
```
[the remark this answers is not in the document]
vpoliakov-pixel: Fixed in 87e8d63 — now appends instead of overwriting: `export JAVA_TOOL_OPTIONS="${JAVA_TOOL_OPTIONS:+${JAVA_TOOL_OPTIONS} }-Djava.awt.headless=true"`, so runtime-provided JVM flags (proxy/TLS/…) are preserved.

plugins/idea/scripts/start-idea.sh
```diff
@@ -65,30 +82,49 @@ else
...
+else
+    pkill -x Xvfb 2>/dev/null || true
+    sleep 0.2
```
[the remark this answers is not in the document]
vpoliakov-pixel: Added in 87e8d63 — the non-headless branch now runs `command -v Xvfb` first and exits with a clear message ("IDEA_HEADLESS=false but Xvfb is not installed … rebuild with Idea(headless=False)") instead of failing with a confusing `command not found`.

plugins/idea/scripts/start-idea.sh
```diff
@@ -65,30 +82,57 @@ else
...
 echo ""
 
-# ── Start IDEA in background ──────────────────────────────────────────────────
-# JAVA_TOOL_OPTIONS is read by the JVM before application startup code, making
-# it the most reliable way to set java.awt.headless=true.
-export JAVA_TOOL_OPTIONS="-Djava.awt.headless=true"
+# ── Start Xvfb (virtual display) unless running headless ──────────────────────
+# IDEA supports -Djava.awt.headless=true natively (the default). When
+# IDEA_HEADLESS=false, Xvfb provides a virtual X11 display that satisfies AWT
+# without real hardware — the same way PyCharm always runs.
```
vfedorishchev: I think we have quite a bit of repetition of the same information across this file to the point, where amount of documentation comes close to the logical lines of the script itself.
It would be great to organise it slightly better, either by using docusaurus, splitting the script or maybe using python instead of bash (BTW is there a good reason to use bash here, like python is not installed at this moment?)

plugins/idea/src/idegym/plugins/idea/resources/Dockerfile.install.j2
```diff
@@ -31,9 +36,49 @@ RUN set -eux; \
...
+# force -Djava.awt.headless=true; virtual-display builds must NOT (they would ignore the
+# Xvfb DISPLAY) and instead suppress the EUA / first-run dialogs, exactly like PyCharm.
 RUN for f in "${IDE_DIR}/bin/idea64.vmoptions" "${IDE_DIR}/bin/idea.vmoptions"; do \
-        [ -f "$f" ] && printf '\n-Djava.awt.headless=true\n-Didea.trust.all.projects=true\n-Djb.consents.confirmation.enabled=false\n-Dide.show.tips.on.startup.default.value=false\n-Dide.no.platform.update=true\n-Dide.browser.jcef.enabled=false\n-XX:-UsePerfData\n-XX:+PerfDisableSharedMem\n-XX:+UseContainerSupport\n-XX:ErrorFile=/tmp/jvm-crash.log\n' >> "$f" || true; \
+        [ -f "$f" ] && printf '\n{% if headless %}-Djava.awt.headless=true\n{% else %}-Djb.privacy.policy.text=<!--999.999-->\n-Dide.firstStartup=false\n{% endif %}-Didea.trust.all.projects=true\n-Djb.consents.confirmation.enabled=false\n-Dide.show.tips.on.startup.default.value=false\n-Dide.no.platform.update=true\n-Dide.browser.jcef.enabled=false\n-XX:-UsePerfData\n-XX:+PerfDisableSharedMem\n-XX:+UseContainerSupport\n-XX:ErrorFile=/tmp/jvm-crash.log\n' >> "$f" || true; \
```
vfedorishchev: probably not a comment to this particular PR, but his lines is extremely difficult to read and validate. Can we split this thing in some sort of an arg array like you did with HEADLESS_ARGS?

plugins/idea/src/idegym/plugins/idea/resources/Dockerfile.install.j2
```diff
@@ -31,9 +36,49 @@ RUN set -eux; \
...
+    printf '%s\n' 'rsch.send.usage.stat:1.1:1:1700000000000' \
+        > /root/.local/share/JetBrains/consentOptions/accepted; \
+    printf '%s\n' \
+        'com.intellij.marketplace' \
+        'com.intellij.settingsSync' \
+        'com.jetbrains.codeWithMe' \
+        'com.intellij.platform.daemon' \
+        > "{{ config_dir }}/disabled_plugins.txt"; \
+    printf '%s\n' \
+        '<application>' \
```
vfedorishchev: works for now,
for the future, it would be great to have a follow-up ticket to adress a couple of issues here.
first thing I see is that we printf the xml into the files just like this, and I would've wished a slightly more elegant solution, which ensures syntax correctness and that editing the resulting file won't corrupt the XML.

Correct me if I'm wrong, but we can probably have another script extracted out of this template, can't we? This might simplify the resulting code a bit, but might come at a cost of the more complex plugin python script.

vfedorishchev (changes requested):

plugins/idea/scripts/start-idea.sh
vpoliakov-pixel: moved a lot of duplicated code and scripts into plugin utils

cannot convert it into python because it should be available for java-only envs too

plugins/idea/src/idegym/plugins/idea/resources/Dockerfile.install.j2
vpoliakov-pixel: introduced HEADLESS_ARGS and split this line into multiline for readability

plugins/idea/src/idegym/plugins/idea/resources/Dockerfile.install.j2
vpoliakov-pixel: refactored this place too

vfedorishchev:
really good job at extracting the shared logic into utils as well as unifying the documentation 👍

vfedorishchev (approved):

PR #275, opened 2026-07-31
[ci] Draft changelog Highlights with Claude Code, anchor the range on CHANGELOG.md

vpoliakov-pixel:
Two changes to the changelog automation.

**1. Highlights are drafted with Claude Code, not in CI.** The `actions/ai-inference` step and the `models: read` permission are gone. The workflow still generates the `CHANGELOG.md` section, syncs the GitHub Release notes, and opens the changelog PR — only the `Highlights` paragraph is deferred. It ships as a placeholder naming the command that fills it in, and the PR body repeats it:

```
_TODO: summarise the headline changes of this release. Draft this paragraph with
`uv run scripts/draft_highlights.py 0.11.0` (uses Claude Code), or write it by hand._
```

New `scripts/draft_highlights.py` chains the two halves the generator already exposed — `--emit-highlights-prompt` out, `--highlights-file` back in — around a one-shot `claude --print`. The drafting instructions are the same prose that used to live in the workflow YAML. It assumes a signed-in `claude` on `PATH`; no API key or token is read. `--sync-release` also pushes the regenerated notes to the GitHub Release via `gh`.

**2. A release section now starts at the last version `CHANGELOG.md` documents, not at the preceding tag.** `v0.11.0` was tagged, failed to publish, and never got a section; `v0.11.1` shipped the fix. Diffing against the preceding *tag* gave `0.11.1` a single bug-fix entry and dropped everything `v0.11.0` carried, permanently. Anchoring on the changelog makes `0.11.1` cover `v0.10.0..v0.11.1` instead — measured on this branch, **1 entry → 96 entries**. The generator writes a line to stderr whenever the two disagree, so a folded-in release is visible in the workflow log; `--previous` still overrides. Documented versions with no matching tag are ignored, since they cannot bound a `git log` range.

## How it works

The drafting step moves out of CI and becomes a maintainer step on the changelog branch.

```mermaid
flowchart TD
  tag["push tag v0.11.0"] --> wf["changelog.yml"]

  subgraph ci["In CI (unattended)"]
    wf --> gen["generate_changelog.py --force"]
    gen --> ph["Highlights placeholder<br/>(was: GitHub Models draft)"]
    ph --> rel["GitHub Release notes"]
    ph --> pr["PR on changelog/v0.11.0"]
  end

  pr -.->|"maintainer, before merging"| dh

  subgraph local["Locally (new)"]
    dh["draft_highlights.py"] --> prompt["generate_changelog.py<br/>--emit-highlights-prompt"]
    prompt --> cc["claude --print"]
    cc --> back["generate_changelog.py<br/>--highlights-file --force"]
  end

  back --> commit["commit CHANGELOG.md"]
  back -->|"--sync-release"| rel
```

Main review risk: the range change alters what a *regenerated* section covers, not just a new one. Re-running the workflow for an already-documented version still diffs against the version below it, so regeneration is idempotent — but `--force` on a version whose predecessor was never documented will now legitimately produce a much larger section.

## Testing

- `uv run pytest -m unit` → **891 passed, 2 skipped, 220 deselected** (871 baseline, +20 new tests across `unit-tests/test_draft_highlights.py` and `test_generate_changelog.py`).
- `uv run ruff check` and `uv run ruff format --check` clean.
- Real end-to-end runs against this repository:
  - `draft_highlights.py 0.11.0 --print` → accurate 4-sentence paragraph in ~8s.
  - `draft_highlights.py 0.11.1 --print` → paragraph covering the OpenHands plugin, build-backend abstraction, pod crash handling and the website, i.e. the `v0.11.0` content that the old range would have lost.
  - Full write path inserts a correct section into `CHANGELOG.md` (reverted afterwards; this PR adds no release content).
- Range detection verified both ways: with `0.11.0` undocumented → 96 entries + the stderr note; with a fixture changelog where `0.11.0` *is* documented → 1 entry, no note.
- Workflow YAML parses, every `run:` block passes `bash -n`, and the PR-body heredoc was rendered with fake env vars to confirm it is not indented into a Markdown code block.

## Follow-up, not fixed here

`[Release]: fix package building` renders as `- : fix package building` — `_LEADING_TAGS_RE` strips the `[Release]` tag but leaves the stray colon. Pre-existing; kept out of this diff.

Also worth deciding separately: whether to regenerate the open `changelog/v0.11.1` PR with this logic so the `0.11.0` changes actually get documented. This PR only changes the tooling.

## Checklist

- [x] [`AGENTS.md`](AGENTS.md) is still accurate — it points at `RELEASING.md` for release conventions, and `RELEASING.md` is updated here.


samorojy (approved):

samorojy:
Why do you support only Claude? :D

PR #278, opened 2026-08-05
[JBRes-9944] Add safe database schema rollback support for IdeGYM deployments

vpoliakov-pixel:
`helm rollback` restores Kubernetes resources and nothing else. The orchestrator migrates the
schema forward on startup, so after a failed upgrade the database is left on the *newer*
revision — which the older image may not contain at all, and Alembic cannot traverse a revision
whose script it does not have.

This adds the explicit, guarded workflow rather than automatic snapshot restoration.

Resolves: JBRes-9944

## What changed

**The bug first.** `001_down.sql` dropped `alembic_version`. Alembic deletes the revision row in
that same transaction, so the statement failed and aborted the whole downgrade — `downgrade base`
was impossible, not merely lossy. Removed on both sides (`001_up.sql` also created the table
redundantly), and a unit test now forbids any migration from naming it.

**Every release declares its schema revision.** `database.schemaRevision` in the chart, rendered
as `IDEGYM_DATABASE_SCHEMA_REVISION` through `required`, so a release cannot omit it. The
orchestrator refuses to start when it disagrees with its own migration head, and a unit test
fails CI when it drifts from Alembic's sole head.

**Exact-target migration.** `MigrationManager.migrate_to(target, allow_downgrade=)` resolves a
`MigrationPlan` before touching anything, then executes it under the existing advisory lock,
re-planning once the lock is held. It refuses — rather than guesses — on an unknown target, a
database sitting on a revision this image does not contain (the signature of rolling back with
the wrong image), a reverted revision that ships no down SQL, a branched history, or any
backwards move without explicit approval. Exposed as `python -m idegym.orchestrator.db_cli`
(`schema current|head|verify`, `migrate --target … [--allow-downgrade] [--dry-run]`).

**`scripts/rollback.py`** compares what the two releases declare. Same revision → plain
`helm rollback --wait`. Older revision → preflight, stop orchestrator *and* watcher, downgrade
with the **currently deployed** image, then Helm, then verify the recorded revision and the
replica counts.

## How it works

The reviewable part is where it stops. Both abort paths leave the release untouched, and a
failed downgrade never reaches Helm — rolling the resources back onto a half-reverted schema
would leave the old code facing a schema it cannot read.

```mermaid
flowchart TD
    A["scripts/rollback.py --revision N"] --> B{"schemaRevision: deployed vs release N"}
    B -->|same| H["helm rollback --wait"]
    B -->|"N declares an older one"| C{"Preflight Job: can the deployed image revert those revisions?"}
    C -->|no| X1["Stop. Nothing was stopped, the release is untouched"]
    C -->|yes| D["Scale orchestrator and watcher to 0"]
    D --> E["Downgrade Job on the deployed image: migrate --target N --allow-downgrade"]
    E -->|failed| X2["Stop. Helm never called, writers stay down, recovery commands printed"]
    E -->|succeeded| H
    H --> V["Verify: schema verify --expect N, both writers ready"]
```

The preflight runs as a Job rather than `kubectl exec` because a rollback usually follows a
failed rollout, when no pod of the source image may be healthy enough to exec into — and a pod of
the *previous* image would answer "can you revert these?" wrongly. It runs exactly the image from
the Deployment spec and is deleted afterwards.

[Database Rollback](https://github.com/JetBrains-Research/idegym/blob/vladimir.poliakov/jbres-9944-db-rollback/website/docs/reference/database_rollback.md)
carries the operator-facing pair: a decision flowchart for which of the three rollbacks applies,
and a phase-labelled sequence diagram of the exact-schema one.

Main review risk: the downgrade Job inherits the live orchestrator's image, env, service account,
tolerations and affinity, but nothing else. A future pod-spec field that migrations depend on
would have to be added to `INHERITED_POD_SPEC_KEYS`.

## Not in this PR

**Backup and restore automation is deliberately deferred** to a follow-up ticket, so two of the
ticket's acceptance criteria (a retained release-bound `pg_dump` backup, a checksum-verified
restore) are not met here. The reason is that `pg_dump` must be at least the server's major
version, which Debian bookworm's `postgresql-client` is not for the bundled PostgreSQL — that
means adding the PGDG repo to the orchestrator image plus hook/PVC templates, a materially larger
and more environment-dependent change than the rollback mechanism itself.
[Database Rollback](https://github.com/JetBrains-Research/idegym/blob/vladimir.poliakov/jbres-9944-db-rollback/website/docs/reference/database_rollback.md)
documents the manual `pg_dump`/`pg_restore` procedure and says plainly that the automation does
not exist yet.

## Testing

- Unit: **930 passed, 2 skipped** (59 new, across `test_db_cli.py`, `test_rollback.py`,
  `test_migrations.py`). The rollback script's cluster calls are mocked; what is pinned down is
  the decision-making and the ordering guarantee.
- DB integration: **94 passed**, including 7 new round-trip tests that drive every revision up
  and down against real PostgreSQL with data seeded at each step, assert the head schema matches
  the ORM models, and cover the `downgrade base` regression. Each test gets a throwaway database
  in the shared container, because Alembic owns the whole schema and cannot share one with the
  ORM-created database the other modules truncate.
- `ruff check` and `ruff format --check` pass; `uv lock --check` unchanged.
- `helm lint` passes, `helm template` renders the new env, and rendering fails as intended when
  `schemaRevision` is null.
- `cd website && npm run build` succeeded when the docs were written; the later diagram edits
  changed only mermaid blocks, so no link or anchor moved (CI's Build site job is the check).

`e2e-tests/test_db_rollback.py` runs in the `other` group. Its first CI run failed all three
tests and was right to: selecting Deployments by `app.kubernetes.io/instance` also matched the
grafana and prometheus subcharts, which share the release, so the rollback aborted before doing
anything on any chart with observability enabled. Fixed in 0e7ede8 by matching on
`app.kubernetes.io/name`, with subchart Deployments added to the unit fixtures.

Note the e2e file cannot stage a literal N → N+1 → N rollback: one image means one set of
migrations, and any running orchestrator drags the schema back to its own head, so no release in
the history can hold it lower. It instead exercises every mechanism that case depends on —
declared revision vs. image vs. database, the downgrade Job built from the live Deployment, a full
down/up round trip with data, and a real application-only `helm rollback` driven through `main()`.

## Checklist

- [x] [`AGENTS.md`](AGENTS.md) is still accurate — the migrations playbook now says a migration is
      a four-file change (module, up SQL, down SQL, `schemaRevision`), that downgrades are executed
      rather than decorative, and how to run the round-trip suite.

### DIFF
--- a/e2e-tests/test_db_rollback.py
+++ b/e2e-tests/test_db_rollback.py
@@ -88,14 +88,12 @@ def psql(sql: str) -> str:
 def cli_output(deployment_name: str, arguments: list[str]) -> str:
     """Run the migration CLI in a live pod and return the value it printed.
 
-    The container logs structured JSON to stdout too, and the command prints its answer
-    last, so the value is the final non-empty line of stdout.
+    The CLI keeps its log records on stderr, so stdout is the answer on its own.
     """
     code, stdout, stderr = exec_in_deployment(deployment_name, DEFAULT_NAMESPACE, [*MIGRATION_CLI, *arguments])
     assert code == 0, f"{' '.join(arguments)} failed: {stderr or stdout}"
-    lines = [line.strip() for line in stdout.splitlines() if line.strip()]
-    assert lines, f"{' '.join(arguments)} printed nothing (stderr: {stderr})"
-    return lines[-1]
+    assert stdout, f"{' '.join(arguments)} printed nothing (stderr: {stderr})"
+    return stdout.strip()
 
 
 def run_migration_job(orchestrator: dict, target: str) -> None:
--- a/e2e-tests/test_db_rollback.py
+++ b/e2e-tests/test_db_rollback.py
@@ -1,15 +1,13 @@
 """End-to-end coverage for release rollback, schema included.
 
-Exercises the two halves of `scripts/rollback.py` against a real cluster and a real
-release: the exact-revision schema move it performs with the deployed image and every
-writer stopped, and the plain application-only rollback it falls back to when the target
-release declares the same schema.
+Exercises both halves of `scripts/rollback.py` against a real cluster: the exact-revision
+schema move, with the deployed image and every writer stopped, and the application-only
+rollback it falls back to when the target release declares the same schema.
 
-The literal N -> N+1 -> N case cannot be staged here, because a single-image cluster has
-only one set of migrations: an orchestrator always migrates to its own head on startup, so
-no release in the history can hold the schema below it. What is covered instead is every
-mechanism that case relies on — the declared revision, the downgrade Job built from the
-live Deployment, the round trip with data in place, and the Helm handover.
+The literal N -> N+1 -> N case cannot be staged here — one image means one set of
+migrations, and any running orchestrator drags the schema back to its own head. Covered
+instead is every mechanism that case relies on: the declared revision, the downgrade Job
+built from the live Deployment, the round trip with data, and the Helm handover.
 """
 
 import json
@@ -97,10 +95,10 @@ def cli_output(deployment_name: str, arguments: list[str]) -> str:
 
 
 def run_migration_job(orchestrator: dict, target: str) -> None:
-    """Run the downgrade/upgrade Job `scripts/rollback.py` would run, and require success.
+    """Run the Job `scripts/rollback.py` would run, and require success.
 
-    ``build_migration_job`` always passes ``--allow-downgrade``; the CLI only consults it
-    when the plan actually moves backwards, so the same builder serves both directions.
+    ``build_migration_job`` always passes ``--allow-downgrade``, which the CLI consults only
+    when the plan moves backwards, so one builder serves both directions.
     """
     name = migration_job_name(orchestrator["metadata"]["name"], f"migrate-{target}")
     delete_job(name, DEFAULT_NAMESPACE)
@@ -143,11 +141,8 @@ def restore_writers(replicas_by_name: dict[str, int]) -> None:
 
 
 def test_release_declares_the_schema_revision_its_image_produces():
-    """The invariant every exact rollback rests on.
-
-    The chart value is what a rollback downgrades to, so a release whose declaration drifts
-    from its image would send the next rollback to the wrong revision.
-    """
+    """The invariant every exact rollback rests on: a declaration that drifts from its image
+    sends the next rollback to the wrong revision."""
     values = release_values(HELM_RELEASE, DEFAULT_NAMESPACE)
     declared = declared_schema_revision(values)
     orchestrator, _ = release_deployments(HELM_RELEASE, DEFAULT_NAMESPACE, values)
@@ -161,9 +156,8 @@ def test_release_declares_the_schema_revision_its_image_produces():
 def test_schema_downgrades_and_comes_back_with_data_intact():
     """Walk the deployed schema back one revision and forward again, in-cluster.
 
-    Rows are seeded into the tables revision 001 creates, so they survive any incremental
-    revision's downgrade; if a future revision's downgrade drops them, that is a
-    data-destroying downgrade this test should fail on.
+    Rows go into the tables revision 001 creates, so they survive an incremental revision's
+    downgrade; one that drops them is data-destroying and should fail here.
     """
     chain = revision_chain()
     assert len(chain) >= 2, "need at least two revisions to exercise a downgrade"
@@ -196,10 +190,9 @@ def test_schema_downgrades_and_comes_back_with_data_intact():
 
 
 def test_application_only_rollback_leaves_the_schema_alone():
-    """A rollback between releases that declare the same revision is a plain Helm rollback.
+    """Same declared revision means a plain Helm rollback: no writer stopped, no Job run.
 
-    No writer is stopped and no migration Job runs — the schema is already what the target
-    release expects, which is the routine case an expand/contract migration policy aims for.
+    This is the routine case an expand/contract migration policy aims for.
     """
     deployed = current_helm_revision(HELM_RELEASE, DEFAULT_NAMESPACE)
     subprocess.run(
--- a/e2e-tests/test_db_rollback.py
+++ b/e2e-tests/test_db_rollback.py
@@ -1,0 +1,249 @@
+"""End-to-end coverage for release rollback, schema included.
+
+Exercises the two halves of `scripts/rollback.py` against a real cluster and a real
+release: the exact-revision schema move it performs with the deployed image and every
+writer stopped, and the plain application-only rollback it falls back to when the target
+release declares the same schema.
+
+The literal N -> N+1 -> N case cannot be staged here, because a single-image cluster has
+only one set of migrations: an orchestrator always migrates to its own head on startup, so
+no release in the history can hold the schema below it. What is covered instead is every
+mechanism that case relies on — the declared revision, the downgrade Job built from the
+live Deployment, the round trip with data in place, and the Helm handover.
+"""
+
+import json
+import subprocess
+
+import pytest
+from alembic.config import Config
+from alembic.script import ScriptDirectory
+from from_root import from_root
+from idegym.utils.logging import get_logger
+from utils import k8s_client
+from utils.constants import (
+    CHART_PATH,
+    DEFAULT_NAMESPACE,
+    HELM_RELEASE,
+    POSTGRESQL_APP_LABEL,
+    POSTGRESQL_DB,
+    POSTGRESQL_USER,
+    WATCHER_APP_LABEL,
+)
+from utils.k8s_setup import wait_for_pod_ready, wait_for_service
+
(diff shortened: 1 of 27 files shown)


scripts/rollback.py
```diff
@@ -0,0 +1,496 @@
...
+def preflight_downgrade(deployment: dict[str, Any], namespace: str, target_revision: str, timeout_seconds: int) -> None:
+    """Ask the source image whether it can reach ``target_revision``, changing nothing.
+
+    Runs as a Job rather than ``kubectl exec`` on purpose: a rollback usually follows a
+    failed rollout, when no pod of the new image may be healthy enough to exec into — and
+    a pod of the *previous* image would answer the question wrongly. The Job runs exactly
+    the image the downgrade will use, resolves the plan, and prints it.
+    """
+    name = f"{deployment['metadata']['name']}-migrate-check-{target_revision}"
+    create_job(build_migration_job(name, namespace, deployment, target_revision, timeout_seconds, dry_run=True))
```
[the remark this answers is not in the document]
vpoliakov-pixel: Fixed in 0e7ede8. Job names now go through `migration_job_name()`, which trims the deployment part so `<deployment>-<suffix>` fits the 63 characters Kubernetes allows (the name becomes a label value on the pods). Both call sites use it, and three unit tests cover the trim, the no-op case, and that trimming cannot leave a double dash.

scripts/rollback.py
```diff
@@ -0,0 +1,496 @@
...
+    )
+
+    print("\nStopping database writers...")
+    for name in original_replicas:
+        scale(name, namespace, 0)
+    for deployment in writers:
+        wait_for_no_pods(deployment, namespace, timeout)
+
+    job_name = f"{orchestrator_name}-migrate-{target_schema}"
+    create_job(build_migration_job(job_name, namespace, orchestrator, target_schema, timeout))
```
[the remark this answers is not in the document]
vpoliakov-pixel: Same fix as the preflight Job above — 0e7ede8 routes this name through `migration_job_name()` too.

e2e-tests/test_db_rollback.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def restore_writers(replicas_by_name: dict[str, int]) -> None:
+    for name, replicas in replicas_by_name.items():
+        scale(name, DEFAULT_NAMESPACE, replicas)
+    assert wait_for_service(), "Orchestrator did not become healthy again"
+    assert wait_for_pod_ready(WATCHER_APP_LABEL, timeout=STEP_TIMEOUT_SECONDS), "Watcher did not become ready again"
+
+
+@pytest.mark.e2e
```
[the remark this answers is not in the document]
vpoliakov-pixel: Agreed, removed in 0e7ede8. AGENTS.md says the same thing (markers are assigned by directory in the root `conftest.py`; do not hand-add them), and only 2 of 20 e2e files had them. Verified `-m e2e` still collects all three tests.

karshkoff (approved):

orchestrator/src/idegym/orchestrator/database/database.py
```diff
@@ -70,10 +70,19 @@ def connect_db_engine(db_url: str, config: SQLAlchemyConfig) -> AsyncEngine:
     return db_engine
 
 
-async def init_db(db_url: str, config: SQLAlchemyConfig, clean_database: bool = False):
+async def init_db(
+    db_url: str,
+    config: SQLAlchemyConfig,
+    clean_database: bool = False,
+    declared_schema_revision: Optional[str] = None,
```
vfedorishchev, line 77: it would be `Optional[str] = None` only to provide backwards compatibility to previous versions of the service, if I got the intent right? If so, how about creating a follow-up ticket making it non-Optional?

orchestrator/src/idegym/orchestrator/db_cli.py
```diff
@@ -0,0 +1,140 @@
...
+from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.ext.asyncio import create_async_engine
+
+logger = get_logger(__name__)
+
+EXIT_OK = 0
+EXIT_ERROR = 1
+
+
+def build_parser() -> ArgumentParser:
```
vfedorishchev, line 58: I'm wondering if we should explore https://typer.tiangolo.com/ for this and other projects where we need to build CLI tools? Seems like a good fit to me
Shouldn't be a priority for this PR for sure

orchestrator/src/idegym/orchestrator/db_cli.py
```diff
@@ -0,0 +1,140 @@
...
+    args = build_parser().parse_args(argv)
+    config = load_config()
+    configure_logging(config=config.logging)
+    configure_sqlalchemy_logging(config=config.logging)
+
+    try:
+        return asyncio.run(_dispatch(args, config))
+    except MigrationError as e:
+        # The message is the operator-facing half of the failure; the traceback is noise.
+        print(f"error: {e}", file=sys.stderr)
```
vfedorishchev, line 98: maybe I'm missing something, but does this `print` end up in logs because we have a logger declared on line 76, or should it be called explicitly?

orchestrator/src/idegym/orchestrator/db_cli.py
```diff
@@ -0,0 +1,140 @@
...
+        manager = MigrationManager(engine=engine, db_url=database.url)
+        if args.command == "schema":
+            return await _run_schema_command(args, manager)
+        return await _run_migrate_command(args, manager)
+    finally:
+        await engine.dispose()
+
+
+async def _run_schema_command(args: Namespace, manager: MigrationManager) -> int:
+    if args.schema_command == "head":
```
vfedorishchev: can we have some consts for all the params we compare against, please?

orchestrator/src/idegym/orchestrator/db_cli.py
```diff
@@ -0,0 +1,140 @@
...
+    if current != args.expect:
+        print(f"error: database is at revision {current}, expected {args.expect}", file=sys.stderr)
+        return EXIT_ERROR
+    return EXIT_OK
+
+
+async def _run_migrate_command(args: Namespace, manager: MigrationManager) -> int:
+    if args.dry_run:
+        plan = manager.plan_migration(current=await manager.get_current_revision(), target=args.target)
+        print(plan.describe())
```
vfedorishchev, line 148: are these prints just debug logs?

orchestrator/src/idegym/orchestrator/migration_manager.py
```diff
@@ -83,23 +140,189 @@ async def _execute_alembic_upgrade(self):
...
+            logger.info("No migration needed", revision=plan.target)
+            return plan
+
+        async with self._advisory_lock() as lock_connection:
+            if lock_connection is None:
+                raise MigrationError(
+                    "Another process holds the migration lock; a migration or rollback is already in progress"
+                )
+
+            # The revision may have moved between planning and taking the lock — a starting
```
vfedorishchev: if revision has been moved it means the new version of orchestrator has been deployed or do you see any other way it can happen? In case that's the only reason it could happen, we should probably cancel the command all together.

website/docs/reference/remote_deployment.md
```diff
@@ -527,6 +527,23 @@ either `helm upgrade` (which restarts pods) or:
...
+orchestrator refuses to start when it disagrees with the migrations in its image.
+
+### Rolling back
+
+`helm rollback` restores Kubernetes resources only; the schema stays at the newer revision.
+Use `scripts/rollback.py`, which resolves the target release's declared revision and downgrades
+the schema first when it differs:
+
+```shell
+uv run python scripts/rollback.py \
```
vfedorishchev, line 540: I think it's fine for now to have a somewhat manual process of rolling back, let's keep an eye on this change to Helm, so that we can hopefully make the rollback automated with helm hooks -> https://github.com/helm/helm/issues/12757

vfedorishchev (approved):

orchestrator/src/idegym/orchestrator/database/database.py
vpoliakov-pixel: JBRes-10375

orchestrator/src/idegym/orchestrator/db_cli.py
vpoliakov-pixel: JBRes-10376

orchestrator/src/idegym/orchestrator/migration_manager.py
vpoliakov-pixel: It already cancels altogether because of an exception.
Deploying new version of an orchestrator should be the reason, but we have multiple replicas of them and all of them try to run migrations

orchestrator/src/idegym/orchestrator/db_cli.py
vpoliakov-pixel: not really, they are results of the commands

PR #280, opened 2026-08-10
[JBRes-9159] Replace Hydra with Pydantic config

samorojy:
Resolves: JBRes-9159

Before: Hydra introduces its own defaults and Pydantic models use their own. etc. 

Now: We assign an ENV name to each field, iterate through the Config Models and collect a list of the ENVs. While loading the config, we iterate through this list and setting chosen ENVs to those from the environ. At the very end we pass it to the Config(**data) which parses it and sets the defaults. 

## Checklist


vpoliakov-pixel:
Is it possible to somehow automate env name based on config class and field name?
AND support previous names for backward compatibility

vpoliakov-pixel:
I think keeping the ability to read config from a file could still be useful to allow configuring service in one place instead of setting hundreds of env vars.
Read defaults from config model, then read values from config file if there is any, then read custom values from env vars

api/src/idegym/api/config.py
```diff
@@ -1,27 +1,65 @@
...
+    host: IPvAddress = Field(default="0.0.0.0", validation_alias="IDEGYM_SERVER_HOST")
+    port: int = Field(ge=0, le=65535, default=8000, validation_alias="IDEGYM_SERVER_PORT")
+    response_buffer_size: MemoryQuantity = Field(
+        ge=0, default=MemoryQuantity(mi=8), validation_alias="IDEGYM_SERVER_RESPONSE_BUFFER_SIZE"
+    )
+    shutdown_delay: Duration = Field(default=Duration(seconds=30), validation_alias="IDEGYM_SERVER_SHUTDOWN_DELAY")
+
+
+class LoggingConfig(ConfigModel):
+    # validate_default: `file_path` is normalised by a validator, which pydantic skips for
```
vpoliakov-pixel: probably could be removed to reduce the amount of code

api/src/idegym/api/config.py
```diff
@@ -33,9 +71,12 @@ def validate_file_path(cls, value: str) -> str:
             return abspath(path)
 
 
-class ProjectConfig(BaseModel):
-    path: str = Field(default=".project")
-    archive: Optional[str] = Field(default=None)
+class ProjectConfig(ConfigModel):
+    # validate_default: see LoggingConfig — `path` and `archive` are normalised by validators.
```
vpoliakov-pixel: probably could be removed to reduce the amount of code

api/src/idegym/api/config.py
```diff
@@ -209,40 +359,52 @@ class MCPConfig(BaseModel):
...
+        validation_alias="IDEGYM_WATCHER_REQUEST_STALE",
     )
 
 
-class OrchestratorConfig(BaseModel):
-    host: IPvAddress = Field(default="0.0.0.0")
-    port: int = Field(ge=0, le=65535, default=8000)
-    workers: int = Field(description="Number of uvicorn worker processes", ge=1, default=1)
+class OrchestratorConfig(ConfigModel):
+    # validate_default: see LoggingConfig — `prometheus_multiproc_dir` is normalised by a validator.
```
vpoliakov-pixel: probably could be removed to reduce the amount of code

backend-utils/src/idegym/backend/utils/settings.py
```diff
@@ -0,0 +1,62 @@
...
+from os import environ
+from typing import Any, Optional
+
+from idegym.api.config import Config
+from pydantic import BaseModel
+
+__all__ = ["ORCHESTRATOR_SECTIONS", "SERVER_SECTIONS", "environment_aliases", "load_config"]
+
+# Which top-level sections each service reads. These replace the per-service Hydra `defaults:`
+ORCHESTRATOR_SECTIONS = frozenset({"logging", "otel", "orchestrator"})
```
vpoliakov-pixel, line 30: do we need to extract watcher section as well?

backend-utils/src/idegym/backend/utils/settings.py
```diff
@@ -0,0 +1,62 @@
...
+SERVER_SECTIONS = frozenset({"logging", "otel", "project", "server"})
+
+
+def environment_aliases(model: type[BaseModel] = Config, prefix: str = "") -> dict[str, str]:
+    """Map every leaf field's dotted path to the environment variable that sets it."""
```
vpoliakov-pixel: would be better to have an explicit example to be able to check the intended logic

vpoliakov-pixel (approved):

backend-utils/src/idegym/backend/utils/settings.py
samorojy: Yes, we do :) 
Will check

backend-utils/src/idegym/backend/utils/settings.py
samorojy: I guess tests should answer your question :)

samorojy:
image

samorojy:
At the moment, we load the orchestrator's config for the watcher because the database interaction logic is coupled with it.

Created a separate ticket for extracting this logic: `JBRes-10405`

samorojy:
```mermaid
flowchart TB
    subgraph declare["Declaration — once per subclass, at import"]
        seg["env_segment"]
        fname["field name"]
        gen["IDEGYM_SEGMENT_FIELD"]
        pin["env(name=…)"]
        old["env(legacy=…)"]
        alias["validation_alias<br/>AliasChoices, ranked"]:::store
        seg --> gen
        fname --> gen
        gen -->|"rank 0"| alias
        pin -->|"replaces rank 0"| alias
        old -->|"rank 1+"| alias
    end

    subgraph load["Load — per call"]
        lc["load_config(sections)"]:::entry
        dv["deprecated_variables(sections)"]:::entry
        guard{"known sections?"}
        err["ValueError"]:::warn
        aliases["environment_aliases()<br/>dotted path → alias list"]
        sel["_selected()<br/>in-scope leaf → first alias set"]
        miss["default kept,<br/>nothing logged"]:::warn
        data["dict keyed by field name"]:::store
        cfg["Config(**data)<br/>populate_by_name, extra=forbid"]:::out
        depr["_deprecated()<br/>winner is not rank 0"]
        logged["one warning per variable"]:::warn
        mapping["legacy → current"]:::out
    end

    alias -.->|"read back"| aliases

    lc --> guard
    dv --> guard
    guard -->|no| err
    guard -->|yes| aliases
    aliases --> sel
    sel -->|"nothing set"| miss
    sel --> data
    sel --> depr
    data --> cfg
    depr -->|"load_config"| logged
    depr -->|"deprecated_variables"| mapping

    classDef entry fill:#6b57ff,stroke:#5b4bd2,color:#fff;
    classDef store fill:#0891b2,stroke:#0e7490,color:#fff;
    classDef warn fill:#e23b3b,stroke:#c02626,color:#fff;
    classDef out fill:#2563eb,stroke:#1d4ed8,color:#fff;

    click alias "https://github.com/JetBrains-Research/idegym/blob/main/api/src/idegym/api/config.py" "ConfigModel and the env helper."
    click sel "https://github.com/JetBrains-Research/idegym/blob/main/backend-utils/src/idegym/backend/utils/settings.py" "environment_aliases and load_config."
```

samorojy:
The next step - stop supporting legacy envs.

vpoliakov-pixel (approved):

PR #284, opened 2026-08-24
[JBRes-10578] Pin mcp-steroid by link and make the unschedulable readiness budget configurable

vpoliakov-pixel:
Two independent fixes from the same ticket, one commit each.

**mcp-steroid pinning.** `0.102.0-r-c68d8f15d` was rejected outright by the version regex, which
allowed a single hex suffix segment. Accepting it is not enough: the derived URL 404s, because
upstream tags do not follow the artifact version — `v0.102` ships
`mcp-steroid-0.102.0-r-c68d8f15d.zip`, while `v0.94.0` ships `mcp-steroid-0.94.0-8682a5ce.zip`. The
tag is now derived from the suffix shape (multi-segment ⇒ two-part tag), and `mcp_steroid_url` pins
an exact link for the next time a release breaks that rule. Default moves to
`0.102.0-r-c68d8f15d`. Revisits #170; the URL is quoted and validated the way #206 does it.

**Unschedulable readiness budget.** `wait_for_pods_ready` failed after 15 consecutive Unschedulable
polls at a hardcoded 2s interval — ~30s, far less than a sysbox or gVisor pool takes to scale up,
and a limit whose meaning moved with the poll rate. It is now a duration in the new
`orchestrator.scheduling` config, and only once that is spent do we read the pods' autoscaler
events: `TriggeredScaleUp`/`Nominated` extends the budget to
`IDEGYM_SCHEDULING_PROVISIONING_TIMEOUT` (15m), anything else keeps the short one and lands in the
error message. The budget never shrinks, so a mid-boot change of mind cannot kill a wait already
6 minutes into a legitimate scale-up. The chart Role gains read-only `events`; without it each wait
warns once and keeps the short budget.

Both budgets normally exceed the caller's `server_start_wait_timeout_in_seconds` (60s by default),
so that timeout is what bounds a start in practice — it now names the last scheduling state
observed, instead of being an unattributable `TimeoutError`.

Resolves: JBRes-10578

## How it works

The unschedulable branch of the poll loop. The autoscaler is consulted only when its answer could
change the outcome, so a healthy wait costs no extra API calls.

```mermaid
flowchart TD
  poll["pods_are_ready() (unchanged)"] --> ready{"ready and nothing terminating?"}
  ready -->|yes| done["return (unchanged)"]
  ready -->|no| unsched{"has_unschedulable_pods?"}
  unsched -->|no| sleep["sleep(scheduling.poll_interval) (was: hardcoded 2s)"]
  unsched -->|yes| spent{"unschedulable_for >= granted? (replaces: 15-consecutive-poll counter)"}
  spent -->|no| sleep
  spent -->|yes| verdict["node_scaling_verdict() (new): read the pods' autoscaler events"]
  verdict -->|"TriggeredScaleUp / Nominated"| grow["granted = provisioning_timeout (new, 15m)"]
  verdict -->|"refused / no event / unreadable"| keep["granted unchanged"]
  grow --> again{"still over budget?"}
  keep --> again
  again -->|no| sleep
  again -->|yes| fail["RuntimeError naming the verdict (new message)"]
  sleep --> poll
```

## Tests

`uv run pytest -m unit` — 1131 passed, 2 skipped. Run repeatedly for `pytest-randomly` order
stability. Where to look:

- `unit-tests/test_kubernetes_client.py` — the budget surviving a mid-wait verdict flip, no verdict
  lookup while inside the budget, `PT0S` staying disabled when a finite budget is on offer,
  unreadable events warning once and keeping the short budget, and a pod-list failure *not* being
  reported as unreadable events.
- `unit-tests/test_image_builder.py` — tag derivation across all six published mcp-steroid releases,
  `mcp_steroid_url` overriding and shell-quoting, non-`http`/non-`.zip` rejection.
- `unit-tests/test_settings.py` — the three new `IDEGYM_SCHEDULING_*` variables and their defaults.

Not run locally (no cluster): `integration`, `e2e` — `e2e-tests/test_{idea,pycharm}_mcp_steroid.py`
are the first real exercise of the 0.102 download. Docs verified with `npm run build`.

## Checklist


vpoliakov-pixel:
### Follow-up (pre-existing, not fixed here): a successful node scale is always reported as a failure

Found while reviewing this PR's blast radius — this branch changed the signature of
`spin_up_or_update_nodes_for_client`, so its callers were re-read. The bug is on `main` already and
is unrelated to this change, so it is noted here rather than fixed in scope.

**Symptom.** Every successful node scale-up for a client with `nodes_count > 0` is logged as
`Failed to release nodes for client <id>` and moves the client to `AvailabilityStatus.DELETION_FAILED`.

**Cause.** `change_number_of_spun_nodes` treats the return value as a success flag:

```python
# orchestrator/src/idegym/orchestrator/nodes_holder.py:301
nodes_released = await spin_up_or_update_nodes_for_client(...)
if not nodes_released:
    had_errors = True
```

`spin_up_or_update_nodes_for_client` has no `return <value>` on any path — it returns `None` on
success — so `not nodes_released` is unconditionally true. The `nodes == 0` branch just above is
fine: `release_nodes_for_client` really does return a bool, which is likely where the pattern was
copied from.

**Impact.** `watcher/src/idegym/watcher/cleanup.py:120-123` consumes that flag:

```python
had_error = await change_number_of_spun_nodes(client_id=client.id, namespace=client.namespace)
if had_error:
    new_status = AvailabilityStatus.DELETION_FAILED
```

So the watcher's cleanup pass mislabels healthy clients as `DELETION_FAILED` and emits a misleading
error log on every tick that scales a client's nodes. Anything downstream that reads that status or
alerts on that message is acting on a value that is always wrong in this path.

**Suggested fix.** Either return `True` from `spin_up_or_update_nodes_for_client` on the success
path (and `False`/raise on failure), or drop the flag and let the existing `except Exception` around
the call be the only failure signal — the `try` block already sets `had_errors = True` on a real
exception, which makes the return value redundant. A unit test asserting
`change_number_of_spun_nodes` returns `False` for a successful scale would pin it.

Happy to file this as its own ticket if it is not already tracked.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -479,51 +484,184 @@ async def deploy_server(
...
+    Fails fast if image pull errors occur `max_image_pull_attempts` times in a row, or if pods stay
+    Unschedulable for longer than the budget their scheduling verdict earns them: pods start on
+    ``scheduling.unschedulable_timeout`` and, once that is spent, an autoscaler that says it is
+    already growing a node for them extends it to ``scheduling.provisioning_timeout``. The budget is
+    a duration rather than a number of polls because what is being waited on — a node booting —
+    takes a time that has nothing to do with how often we look, and it only ever grows: an
+    autoscaler that will not widen the pool says nothing about capacity a finishing pod is about to
+    free, so a refusal sharpens the error without shortening the wait.
     Raises asyncio.TimeoutError if `wait_timeout` seconds elapse without all pods becoming ready.
     """
```
samorojy, line 521:
```suggestion
    Fails fast if image pull errors occur `max_image_pull_attempts` times in a row, or if pods stay
    Unschedulable for longer than the budget their scheduling verdict earns them: pods start on
    ``scheduling.unschedulable_timeout`` and, once that is spent, an autoscaler that says it is
    already growing a node for them extends it to ``scheduling.provisioning_timeout``.
   Raises asyncio.TimeoutError if `wait_timeout` seconds elapse without all pods becoming ready.
   """
```

samorojy: Too big comment with too AI explanation for my taste :)

samorojy (approved):
Well done!

Could you fix the follow-up in the same request (or stacked)? I guess there should be a single method alternation with correct return on a happy-path.

PR #285, opened 2026-08-25
[JBAI-21836] Improve sandbox placement locality

igoldov:
Resolves: JBAI-21836

## Operational context

Our agentic RL workload starts concurrent fleets of short-lived sandboxes, with several rollouts often using the same multi-GB image. During a cold burst, Kubernetes ImageLocality has no cached-image signal yet; `nodes_count` only keeps capacity alive with pause pods and neither pre-pulls the task image nor assigns sandbox pods to those nodes. The result can be duplicate pulls of one image across several nodes and slower, less predictable group startup.

This PR addresses only that cold-start window: when explicitly enabled, it adds a soft same-image-reference preference. It does not create exact GRPO placement groups

## Summary

- Add an opt-in preferred pod-affinity policy for simultaneous sandbox cold starts using the same image reference.
- Keep the policy disabled by default and configure its 1-100 weight through orchestrator environment settings.
- Preserve existing node, pod, and anti-affinity rules while appending the same-image preference.
- Use the shared SHA-256 helper to produce a compact Kubernetes label value from the exact image reference.
- Document namespace scope, tag-versus-digest behavior, native Kubernetes ImageLocality, and the scheduling tradeoffs.

## Design rationale

Kubernetes ImageLocality handles warm node caches. In the initial cold tie, this optional policy lets later scheduled pods prefer the first pod's host and share one pull instead of starting duplicate pulls across nodes.

This intentionally favors packing over even spreading when enabled. It remains a soft preference, is scoped to one namespace, and should stay disabled where inter-pod-affinity scheduler cost or even distribution is more important.

## Configuration

- `IDEGYM_SAME_IMAGE_AFFINITY_ENABLED=False`
- `IDEGYM_SAME_IMAGE_AFFINITY_PREFERENCE_WEIGHT=100`

The hash identifies the textual image reference, not resolved image bytes. Mutable tags such as `:latest` therefore retain the same label when their contents change; immutable references are recommended when that distinction matters.

## Relationship to the other PRs

The three PRs are independent and address different phases of the same sandbox workload: [#285](https://github.com/JetBrains-Research/idegym/pull/285) improves cold-image placement, [#286](https://github.com/JetBrains-Research/idegym/pull/286) bounds Bash result retention and hardens process cleanup inside a running sandbox, and [#287](https://github.com/JetBrains-Research/idegym/pull/287) enforces a hard sandbox-density ceiling per node.

If #285 and #287 are enabled together, #285 may pack same-image pods only until #287's scheduler-enforced cap is reached; further pods spill to another eligible node.

## Validation

- [x] Focused placement/config/router tests: 188 passed
- [x] Full unit suite: 1,097 passed, 2 skipped
- [x] Repository-wide Ruff format and lint checks
- [x] `uv lock --check`
- [x] Docusaurus production build
- [x] `git diff --check`
- [x] Independent final review: no findings

### DIFF
--- a/backend-utils/src/idegym/backend/utils/kubernetes_client.py
+++ b/backend-utils/src/idegym/backend/utils/kubernetes_client.py
@@ -15,6 +15,7 @@ from idegym.api.status import Status
 from idegym.api.type import ConditionStatus
 from idegym.utils.dict import deep_merge
 from idegym.utils.functools import cached_async_result
+from idegym.utils.hashing import sha256
 from idegym.utils.logging import get_logger
 from kubernetes_asyncio.client import (
     ApiClient,
@@ -51,6 +52,8 @@ from kubernetes_asyncio.client import (
     V1ObjectMeta,
     V1OwnerReference,
     V1Pod,
+    V1PodAffinity,
+    V1PodAffinityTerm,
     V1PodDisruptionBudget,
     V1PodDisruptionBudgetList,
     V1PodDisruptionBudgetSpec,
@@ -71,6 +74,7 @@ from kubernetes_asyncio.client import (
     V1Toleration,
     V1Volume,
     V1VolumeMount,
+    V1WeightedPodAffinityTerm,
 )
 from kubernetes_asyncio.config import (
     ConfigException,
@@ -83,6 +87,37 @@ KubernetesV1Apis = tuple[AppsV1Api, BatchV1Api, CoreV1Api, PolicyV1Api, CustomOb
 V1ResourceList = V1ConfigMapList | V1DeploymentList | V1PodDisruptionBudgetList | V1ServiceList
 
 logger = get_logger(__name__)
+
+SANDBOX_IMAGE_LABEL = "idegym.jetbrains.com/image-id"
+
+
+def sandbox_image_id(image_tag: str) -> str:
+    """Hash the exact image reference string, not its registry-resolved digest."""
+    return sha256(image_tag)[:32]
+
+
+def _add_same_image_affinity(affinity: Optional[V1Affinity], image_id: str, preference_weight: int) -> V1Affinity:
+    """Append same-image preference without replacing existing affinity rules."""
+    affinity = affinity or V1Affinity()
+    pod_affinity = affinity.pod_affinity or V1PodAffinity()
+    preferred = list(pod_affinity.preferred_during_scheduling_ignored_during_execution or ())
+    preferred.append(
+        V1WeightedPodAffinityTerm(
+            weight=preference_weight,
+            pod_affinity_term=V1PodAffinityTerm(
+                label_selector=V1LabelSelector(
+                    match_labels={
+                        "app.kubernetes.io/component": "sandbox",
+                        SANDBOX_IMAGE_LABEL: image_id,
+                    }
+                ),
+                topology_key="kubernetes.io/hostname",
+            ),
+        )
+    )
+    pod_affinity.preferred_during_scheduling_ignored_during_execution = preferred
+    affinity.pod_affinity = pod_affinity
+    return affinity
 
 
 def build_node_affinity(taint_key: str, preference_weight: int) -> V1NodeAffinity:
@@ -264,6 +299,8 @@ async def deploy_server(
     server_kind: ServerKind = ServerKind.IDEGYM,
     snapshot_id: Optional[str] = None,
     snapshot_tag: Optional[str] = None,
+    same_image_affinity_enabled: bool = False,
+    same_image_affinity_preference_weight: int = 100,
 ):
     """
     Create a Kubernetes Deployment, Service, and PodDisruptionBudget for a server.
@@ -340,6 +377,8 @@ async def deploy_server(
         "app.kubernetes.io/version": __version__,
         "idegym.jetbrains.com/snapshot-id": snapshot_id or server_name,
     }
+    if same_image_affinity_enabled:
+        labels[SANDBOX_IMAGE_LABEL] = sandbox_image_id(image_tag)
 
     toleration = (
         V1Toleration(
@@ -359,6 +398,13 @@ async def deploy_server(
         if node_pool_taint_key
         else None
     )
+
+    if same_image_affinity_enabled:
+        affinity = _add_same_image_affinity(
+            affinity,
+            image_id=labels[SANDBOX_IMAGE_LABEL],
+            preference_weight=same_image_affinity_preference_weight,
+        )
 
     pod_spec = V1PodSpec(
         containers=[container],
--- a/unit-tests/test_kubernetes_client.py
+++ b/unit-tests/test_kubernetes_client.py
@@ -8,7 +8,14 @@ only for its (offline) camelCase deserializer.
 import pytest
 from idegym.api.orchestrator.servers import StartServerRequest
 from idegym.backend.utils import kubernetes_client as kc
-from kubernetes_asyncio.client import ApiClient
+from kubernetes_asyncio.client import (
+    ApiClient,
+    V1Affinity,
+    V1LabelSelector,
+    V1PodAffinity,
+    V1PodAffinityTerm,
+    V1WeightedPodAffinityTerm,
+)
 
 
 @pytest.fixture
@@ -56,6 +63,7 @@ async def test_deploy_without_overrides_leaves_pod_unchanged(mocker, api_client)
     assert pod.volumes is None
     assert container.volume_mounts is None
     assert container.env_from is None
+    assert pod.affinity is None
 
 
 async def test_deploy_applies_typed_volumes_mounts_and_env_from(mocker, api_client):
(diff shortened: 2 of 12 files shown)


website/docs/reference/remote_deployment.md
```diff
@@ -343,7 +343,12 @@ The Docker config should contain credentials for your registry:
 
 ---
 
-## Step 9: Dedicated Node Pools (Optional)
+## Step 9: Sandbox Placement and Dedicated Node Pools (Optional)
+
+Sandbox pods automatically prefer nodes in the same namespace that already run the same
+image. IdeGYM uses a hash of the image reference as a pod label and preferred hostname
+affinity, so the registry path is not exposed in metadata and capacity constraints may still
```
vpoliakov-pixel: We already have image tag in pods specs and we you can read pod labels, then most likely you can read pod specs, so the wording that hashing is safer because it does not expose registry is not justified.
I would phrase it somewhat like (do not quote though) "shorter and could be used as a label because image tags have forbidden symbols like : inside"

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -84,6 +88,13 @@
 
 logger = get_logger(__name__)
 
+SANDBOX_IMAGE_LABEL = "idegym.jetbrains.com/image-id"
+
+
+def sandbox_image_id(image_tag: str) -> str:
+    """Return a stable, non-sensitive Kubernetes label value for an image reference."""
+    return sha256(image_tag.encode()).hexdigest()[:32]
```
vpoliakov-pixel: Worth mentioning that we hash the tag and not the digest, so images with `:latest` will have the same hash.

And we also have hashing.py in common-utils for hashing, so it would be great to reuse that

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -360,6 +372,24 @@ async def deploy_server(
         else None
     )
 
+    affinity = affinity or V1Affinity()
+    affinity.pod_affinity = V1PodAffinity(
```
vpoliakov-pixel: overwrite vs merge
We are slowly extending IdeGYM with more configuration functionality so it would be great to have a merging pod_affinity code here instead of overwriting

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -360,6 +372,24 @@ async def deploy_server(
...
+        preferred_during_scheduling_ignored_during_execution=[
+            V1WeightedPodAffinityTerm(
+                weight=100,
```
vpoliakov-pixel: Currently it adds this affinity for every pod, I would prefer seeing this functionality on demand, guarded by some configs. It needs a config flag and a weight parameter similar to `node_pool_preference_weight`

unit-tests/test_kubernetes_client.py
```diff
@@ -102,6 +102,54 @@ async def test_pod_overrides_concatenate_lists_keeping_managed_entries(mocker, a
     assert toleration_keys == {"node-pool", "dedicated"}
 
 
+async def test_same_image_prefers_same_node(mocker, api_client):
```
vpoliakov-pixel, line 113: Now we do not have a test without `node_pool_taint_key`
We have test_deploy_without_overrides_leaves_pod_unchanged but it does not check affinity which is not none anymore

unit-tests/test_kubernetes_client.py
```diff
@@ -102,6 +102,54 @@ async def test_pod_overrides_concatenate_lists_keeping_managed_entries(mocker, a
...
+                            },
+                        }
+                    ]
+                }
+            }
+        },
+    )
+
+    preferred = pod.affinity.pod_affinity.preferred_during_scheduling_ignored_during_execution
+    assert [term.weight for term in preferred] == [100, 10]
```
vpoliakov-pixel: assert sorted(term.weight for term in preferred) == [10, 100]

vpoliakov-pixel:
>The unsafe logical-slot hard cap was removed. It could strand capacity after churn and deadlock Deployment rolling updates.

No code related to this

>Placement/config tests: 165 passed

Not sure which tests were run

vpoliakov-pixel:
I assume that the goal of this change is "to avoid downloading the same multi-GB images multiple times"?

If that is true, then I would suggest to tackle it from a different angle.

1. Each idegym client during registration can specify the number of nodes it needs. The logic is within nodes_holder.py file. Orchestrator will start simple pause containers with specific pdbs and nodes will live until the client is stopped.
2. Kubernetes already has ImageLocality plugin for scheduler by default: Favors nodes that already have the container images that the Pod runs. So each node will download the image only once AND kubernetes will automatically prefer nodes with those images already available in cache.

Even though the current implementation achieves the image locality but it has several limitations:
1. If a client has `nodes_count=4` with this affinity enabled, the nodes would be populated one by one to the fullest, instead of spreading the load evenly
2. Also all consequent pods on the same node will wait until the first pod will download the image
3. As a warning: the whole approach uses InterPodAffinity which is not a good solution for situations when you have either a lot of nodes or bursts with a lot of pods because it triggers a lot of internal cross check requests (using node affinity and image locality is the recommended way still)

unit-tests/test_kubernetes_client.py
igoldov: Done in `215d164`. The default/no-taint path now asserts that affinity remains `None` and the image label is absent. The enabled no-taint test verifies pod affinity is added while node affinity remains absent; composition with node-pool affinity is covered separately.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
igoldov: Done in `215d164`. `_add_same_image_affinity` retains existing node affinity, pod anti-affinity, required pod-affinity terms, and preferred terms, then appends the managed preference. A direct merge regression test and the later `pod_overrides` composition test cover both paths.

website/docs/reference/remote_deployment.md
igoldov: Done in `215d164`. The docs now describe the hash as a compact, label-safe representation because image references contain characters such as `/`, `:`, and `@`; the unsupported registry-secrecy claim is removed. They also state the namespace scope.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
igoldov: Done in `215d164`. SHA-256 now lives in `idegym.utils.hashing` and the placement code reuses it. The docs and helper explicitly say the exact reference string is hashed, not a registry-resolved digest, so a mutable `:latest` reference keeps the same hash.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
igoldov: Done in `215d164`. The behavior is disabled by default and guarded by `IDEGYM_SAME_IMAGE_AFFINITY_ENABLED`; `IDEGYM_SAME_IMAGE_AFFINITY_PREFERENCE_WEIGHT` configures a validated 1-100 weight. Both fresh-server and snapshot deployment paths propagate the settings, with tests.

unit-tests/test_kubernetes_client.py
igoldov: Done in `215d164`: the composition assertion is now order-independent: `sorted(term.weight for term in preferred) == [10, 100]`.

igoldov:
Addressed the PR-description issue from https://github.com/JetBrains-Research/idegym/pull/285#issuecomment-5437756169: the unrelated hard-cap paragraph is removed, and the validation section now names the exact focused and full-unit suites with their pass/skip counts, plus each repository check that was run.

igoldov:
On the architectural point from https://github.com/JetBrains-Research/idegym/pull/285#issuecomment-5437955695: yes, the target is simultaneous same-image cold bursts. I agree that native ImageLocality is the right mechanism once an image is already cached. The gap here is the initial cold tie: nodes_count starts pause holders but neither binds sandbox Deployments to those nodes nor pre-pulls sandbox images, so ImageLocality has no distinguishing cache signal yet.

When explicitly enabled, packing later same-reference pods near the first scheduled pod—and sharing its in-progress pull—is intentional. The implementation is now disabled by default, exposes a configurable preference weight, and documents both InterPodAffinity scaling cost and the loss of even spreading. This remains a soft, namespace-scoped same-reference preference; it is not a general allocator for exact GRPO placement groups.

vpoliakov-pixel:
The code looks okay now, with all the introduced fixes for the comments. But I still have the architectural concerns because packing pods node by node contradicts with node holder functionality and has some scalability problems related to InterPodAffinity. There should be a better way of pre-caching images per node before the run rather than using pod affinity to indirectly coordinate image scheduling and pulling e.g. image puller pods or daemon sets

PR #286, opened 2026-08-25
[JBAI-21836] Bound and sanitize bash executor output

igoldov:
Resolves: JBAI-21836

## Operational context

Agentic rollouts execute model-authored Bash commands whose stdout or stderr can be unexpectedly large, and timed-out commands can leave descendant processes holding inherited output pipes. Retaining complete streams increases server and client memory use and inflates returned, persisted, and logged payloads; lingering pipes can turn a command timeout or cancellation into a rollout stall. 

This PR bounds retained output per stream and makes timeout and cancellation cleanup deterministic. It still drains produced output until command completion or timeout to avoid pipe deadlocks, so it is not a kill-after-total-bytes policy. Parser-sensitive reward checks explicitly opt into complete output.

## Summary

- Add configurable `max_output_bytes` propagation from the Python client and orchestrator MCP through the HTTP tool to `BashExecutor`.
- Retain bounded head-and-tail stdout and stderr with a deque-based collector while tracking total produced bytes.
- Replace malformed UTF-8, preserve leading indentation, and emit bounded structured log excerpts.
- Preserve partial output and byte totals in timeout logs.
- Bound timeout/cancellation pipe draining, close inherited pipes, reap the process, and escalate surviving process-group members to `SIGKILL`.
- Keep parser-sensitive `TestChecker` calls explicitly unlimited and document client, MCP, tool, and architecture behavior.

## Semantics

- External Bash tool calls retain 1 MiB per stream by default.
- Callers may choose another positive per-stream byte limit or use `None` / JSON `null` for trusted output that must remain complete.
- Zero, negative, boolean, and non-integer limits are rejected.
- The truncation marker is outside the configured retained-byte limit.
- Produced output is still drained until command completion or execution timeout to prevent pipe deadlocks. A separate kill-after-total-produced-bytes ceiling remains follow-up work.

## Relationship to the other PRs

The three PRs are independent and address different phases of the same sandbox workload: [#285](https://github.com/JetBrains-Research/idegym/pull/285) concerns cold-image pod placement, [#286](https://github.com/JetBrains-Research/idegym/pull/286) concerns one Bash process and its output after a sandbox is running, and [#287](https://github.com/JetBrains-Research/idegym/pull/287) concerns the maximum number of sandboxes schedulable on one node.

#286 neither changes Kubernetes placement nor imposes a sandbox-capacity policy.

## Validation

- [x] Full unit suite: 1,109 passed, 2 skipped
- [x] Targeted tool, reward-parser, and real-process integration suite: 31 passed
- [x] Repository-wide Ruff format and lint checks
- [x] `uv lock --check`
- [x] Docusaurus production build
- [x] Generated server OpenAPI matches the committed schema
- [x] `git diff --check`
- [x] Independent final review, including detached pipe holders, cancellation, zero grace, signal races, and `SIGTERM`-ignoring process-group children

### DIFF
--- a/backend-utils/src/idegym/backend/utils/bash_executor.py
+++ b/backend-utils/src/idegym/backend/utils/bash_executor.py
@@ -3,6 +3,7 @@ import os
 import shlex
 import signal
 from asyncio.subprocess import Process
+from collections import deque
 from importlib.resources import files
 from pathlib import Path
 from typing import Optional
@@ -14,6 +15,11 @@ from idegym.utils.logging import get_logger
 logger = get_logger(__name__)
 
 __BASH_INIT_FILEPATH__ = files(resources).joinpath("bash-integration.bash")
+_READ_CHUNK_BYTES = 64 * 1024
+_LOG_EXCERPT_CHARS = 1800
+_OUTPUT_DRAIN_TIMEOUT_SECONDS = 0.25
+_PROCESS_GROUP_POLL_INTERVAL_SECONDS = 0.01
+_PROCESS_REAP_TIMEOUT_SECONDS = 0.25
 
 
 class BashExecutorError(Exception):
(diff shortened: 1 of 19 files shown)


backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -14,6 +14,9 @@
 logger = get_logger(__name__)
 
 __BASH_INIT_FILEPATH__ = files(resources).joinpath("bash-integration.bash")
+_MAX_CAPTURE_BYTES_PER_STREAM = 1024 * 1024
```
vpoliakov-pixel: It would be very useful to have it as a parameter that comes all the way from idegym-client -> orchestator -> forward request -> bash tool -> bash_executor, similar to timeout and graceful_termination_timeout.
I can imagine situations when someone needs more output from commands to be logged and someone less.
It would also eliminate the monkeypatching of global vars in unit tests.

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -14,6 +14,9 @@
...
+_READ_CHUNK_BYTES = 64 * 1024
+_LOG_EXCERPT_CHARS = 4000
```
vpoliakov-pixel: It could be confusing to mix the units of data for these parameters (we measure capturing outputs in bytes and cutting log messages in chars) but I am okay with leaving it as is for now.
I would suggest using lower value here, e.g. 1800 aka ~1 page of text.

backend-utils/src/idegym/backend/utils/bash_executor.py
vpoliakov-pixel: MCP tool path would also be affected here, and probably some unit tests that do not use keyword args.

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -71,19 +125,28 @@ async def execute_bash_command(
         )
 
         try:
-            stdout_bytes, stderr_bytes = await asyncio.wait_for(process.communicate(), timeout=timeout)
+            stdout_bytes, stderr_bytes, stdout_total, stderr_total = await asyncio.wait_for(
+                _communicate_bounded(process, _MAX_CAPTURE_BYTES_PER_STREAM), timeout=timeout
```
vpoliakov-pixel: `process.communicate()` closes the output pipes after reading them, `_communicate_bounded` drains them manually without closing explicitly, it would be great to close them in a finally block to avoid leaking resources

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -24,6 +27,57 @@ class BashCommandExecutionTimeoutError(BashExecutorError):
...
+            tail.extend(chunk)
+            overflow = len(tail) - tail_limit
+            if overflow > 0:
+                del tail[:overflow]
+
+    if total <= max_bytes:
+        return bytes(head + tail), total
+    omitted = total - len(head) - len(tail)
+    marker = f"\n... [IdeGYM truncated {omitted} output bytes] ...\n".encode()
+    return bytes(head) + marker + bytes(tail), total
```
vpoliakov-pixel: There are a couple of inconsistencies with this approach:
1) If max_bytes is 0, then it would be only `marker` and this could confuse the end user
2) I would suggest having None as a no-limit
3) Some already existing code rely on output parsing e.g. `check_repository_tests` in `TestChecker` and it is a yet another reason to have it as a parameter (check_repository_tests would not work well with marker line)

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -24,6 +27,57 @@ class BashCommandExecutionTimeoutError(BashExecutorError):
     pass
 
 
+async def _read_bounded(stream: asyncio.StreamReader, max_bytes: int) -> tuple[bytes, int]:
```
vpoliakov-pixel: This method would solve only half of a problem - it would not have long outputs in memory, it would not propagate long outputs back to the client and would not write them in a database but if an agent is stuck and produces enormous amount of output we would still read it till the end. Could be worth having some other hard ceiling limits to not allow commands to produce outputs "till timeout".
I am okay with fixing it as a follow up.

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -71,19 +125,28 @@ async def execute_bash_command(
...
-            stdout_text = stdout_bytes.decode("utf-8").strip() if stdout_bytes else ""
-            stderr_text = stderr_bytes.decode("utf-8").strip() if stderr_bytes else ""
+            stdout_text = stdout_bytes.decode("utf-8", errors="replace").rstrip() if stdout_bytes else ""
+            stderr_text = stderr_bytes.decode("utf-8", errors="replace").rstrip() if stderr_bytes else ""
 
             exit_code = process.returncode
 
         except TimeoutError:
-            logger.warning(f"Command execution timed out after {timeout} seconds")
+            logger.warning("Command execution timed out", timeout_seconds=timeout)
```
vpoliakov-pixel: Now we have dedicated output buffers and can log them here as well in case of timeouts.
Could you please add that?

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -24,6 +27,57 @@ class BashCommandExecutionTimeoutError(BashExecutorError):
...
+        total += len(chunk)
+        head_space = head_limit - len(head)
+        if head_space > 0:
+            head.extend(chunk[:head_space])
+            chunk = chunk[head_space:]
+        if chunk and tail_limit > 0:
+            tail.extend(chunk)
+            overflow = len(tail) - tail_limit
+            if overflow > 0:
+                del tail[:overflow]
```
vpoliakov-pixel: Constant deleting and extending array could cause a lot of data moving around in memory due to compaction e.g. for reading 1 gb of output using 64kb chunks to collect 1 mb of data from both ends it could use several gb of RAM.
Instead of del/extend I would suggest using lists/deque of chunks and merge them into one string only in the end. Adding/removing elements in deque is fast and does not require compaction

unit-tests/test_bash_executor.py
```diff
@@ -0,0 +1,64 @@
+import pytest
+from idegym.backend.utils import bash_executor
+
+
+@pytest.fixture(autouse=True)
```
vpoliakov-pixel: I would suggest testing _read_bounded in unit tests and testing actual bash command in test_bash_executor in docker_tests to avoid spawning subprocesses in unit tests

website/docs/architecture/rewards-tools.md
```diff
@@ -53,7 +53,7 @@ Every server ships the **tools** plugin, available on all images:
 
 | Tool | Endpoint | What it does |
 |---|---|---|
-| Bash | `POST /api/tools/bash` | Run a bash script; returns stdout/stderr/exit code |
+| Bash | `POST /api/tools/bash` | Run a bash script; returns bounded stdout/stderr and the exit code |
```
vpoliakov-pixel, line 56: Need to update website/docs/reference/tools.md and MCP run_bash_command docs, especially after introducing new parameter

vpoliakov-pixel:
>Removes the unreferenced runtime-overlay Dockerfile.

Not sure what it means, no code related to that

backend-utils/src/idegym/backend/utils/bash_executor.py
igoldov: Done in `cdce3cf`. Output draining now has explicit final cleanup: it gets a bounded drain window, then lingering stdout/stderr transports are closed and reader tasks are cancelled and awaited. Process reaping is bounded as well. Real-process tests cover detached descendants retaining pipes, zero grace, cancellation, and a process-group child ignoring `SIGTERM`.

backend-utils/src/idegym/backend/utils/bash_executor.py
igoldov: Done in `cdce3cf`. The collector keeps a fixed head and a deque-backed chunk tail, trimming only boundary chunks and joining once when output is requested. This removes repeated front compaction of a large `bytearray`.

backend-utils/src/idegym/backend/utils/bash_executor.py
igoldov: Done in `cdce3cf`. `max_output_bytes` now flows from both public Python client entry points and the orchestrator MCP request through forwarding, the server request model, tool service, and `BashExecutor`. The new public parameters are appended, so existing positional callers keep their meaning; propagation and MCP serialization have regression coverage.

backend-utils/src/idegym/backend/utils/bash_executor.py
igoldov: Done in `cdce3cf`: structured command/output excerpts are now limited to 1,800 characters, independently of the byte-based retained-output limit.

website/docs/architecture/rewards-tools.md
igoldov: Done in `cdce3cf`. The Bash tool, Python client, orchestrator MCP, and architecture references now document `max_output_bytes`, the 1 MiB default, `null` unlimited mode, head/tail markers, and continued draining. The committed server OpenAPI schema was regenerated and verified.

backend-utils/src/idegym/backend/utils/bash_executor.py
igoldov: Done in `cdce3cf`. Timeout logs now include retained partial stdout/stderr plus the total byte count observed for each stream, after bounded shutdown draining completes.

backend-utils/src/idegym/backend/utils/bash_executor.py
igoldov: Done in `cdce3cf`. The limit is strict positive-integer-or-null: zero, negatives, booleans, and non-integers are rejected; `None` / JSON `null` means unlimited. External calls default to 1 MiB per stream, while parser-sensitive `TestChecker` commands explicitly request unlimited output. All paths have coverage.

unit-tests/test_bash_executor.py
igoldov: Done in `cdce3cf`. Collector behavior and validation are covered by unit tests without spawning subprocesses; real Bash behavior, decoding, truncation, timeouts, cancellation, and process cleanup live under `integration-tests/docker_tests/test_bash_executor.py`.

backend-utils/src/idegym/backend/utils/bash_executor.py
igoldov: Agreed. This PR deliberately bounds retained, returned, and logged output but continues draining produced bytes until completion or the execution timeout to avoid pipe deadlocks. A kill-after-total-produced-bytes ceiling remains separate follow-up work; this distinction is now explicit in the docs and PR description.

igoldov:
The stale runtime-overlay Dockerfile statement has been removed from the PR description. No Dockerfile change is part of this PR.

vpoliakov-pixel (approved):

PR #289, opened 2026-08-31
[JBRes-10676] Bash tool: verbatim output and safe logging (stack 1/8)

vpoliakov-pixel:
**Stack 1/8** — **#289** ← #290 ← #291 ← #292 ← #293 ← #294 ← #295 ← #296 (merge in order, oldest first).
Base: `main`. Branch: `vladimir.poliakov/jbres-10676-01-…`.

JBRes-10676 splits into 18 independent improvements found while integrating IdeGYM as a NeMo Gym sandbox provider. One PR would be 59 commits, so it is a stack, cut by subsystem and ordered so each layer only depends on the ones below. Every branch is independently green; **merge the whole stack in order** rather than piecemeal.

This one is the two High-priority bash-tool fixes. Depends on nothing.

## What changed

**Output is returned verbatim** (improvement #1). The executor trimmed every command's output, so a caller could not tell `printf 'x'` from `printf '  x  '` and a trailing newline that is part of a `git diff` was dropped. `BashCommandRequest.strip_output` is the opt-in for the old behaviour.

**Command and output logging moved to DEBUG** (#2). Both went to the log at INFO. Since setting per-command environment means writing `export TOKEN=…` into the script, that leaked credentials verbatim; and because file transfer rides base64 over bash, a 10 MB upload wrote hundreds of megabytes of logs. INFO now carries only command length, exit code and per-stream byte counts. The command moved to DEBUG behind a redaction pass that masks every `export NAME=…` value.

## How to verify

```
uv run pytest -m unit unit-tests/test_bash_executor.py
uv run pytest integration-tests/docker_tests/test_bash_executor.py
```

- `unit-tests/test_bash_executor.py` — decode/strip behaviour, the redaction table, redact-before-truncate.
- `integration-tests/docker_tests/test_bash_executor.py` — real bash: padded and non-UTF-8 output round-trips with its exit code; an exported credential reaches neither the INFO records nor the DEBUG excerpt.

Full suite on this branch: **1173 passed, 2 skipped**; `ruff check` clean; docs site builds.

## Breaking change

Output is no longer stripped. Every caller in this repository and in the e2e suite already strips its own, but downstream consumers comparing exact output should pass `strip_output=True`.

Resolves: JBRes-10676 (partially — #1, #2)

### DIFF
--- a/integration-tests/docker_tests/test_bash_executor.py
+++ b/integration-tests/docker_tests/test_bash_executor.py
@@ -185,9 +185,40 @@ class TestBashExecutor:
             )
 
         timeout_log = next(entry for entry in logs if entry["event"] == "Command execution timed out")
-        assert timeout_log["stdout"] == "partial output"
+        assert timeout_log["log_level"] == "warning"
         assert timeout_log["stdout_bytes"] == len("partial output")
-        assert timeout_log["stderr"] == ""
+        assert "stdout" not in timeout_log
+
+        partial_log = next(entry for entry in logs if entry["event"] == "Partial output of the timed-out command")
+        assert partial_log["log_level"] == "debug"
+        assert partial_log["stdout"] == "partial output"
+        assert partial_log["stderr"] == ""
+
+    @pytest.mark.asyncio
+    async def test_info_logs_carry_no_command_text_or_output(self):
+        executor = BashExecutor()
+
+        with capture_logs() as logs:
+            await executor.execute_bash_command("export TOKEN=s3cr3t; printf 'result'")
+
+        info_logs = [entry for entry in logs if entry["log_level"] == "info"]
+        assert "s3cr3t" not in repr(info_logs)
+        assert "result" not in repr(info_logs)
+
+        completed = next(entry for entry in info_logs if entry["event"] == "Command completed")
+        assert completed["exit_code"] == 0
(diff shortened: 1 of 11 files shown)


integration-tests/docker_tests/test_bash_executor.py
```diff
@@ -112,6 +112,30 @@ async def test_exit_command(self):
...
+
+        assert stdout == "padded"
+        assert exit_code == 0
+
+    @pytest.mark.asyncio
+    async def test_non_utf8_output_is_replaced_and_exit_code_survives(self):
+        executor = BashExecutor()
+        stdout, _stderr, exit_code = await executor.execute_bash_command("printf '\\xff\\xfe'; exit 3")
+
+        assert stdout == "��"
```
samorojy: ??

integration-tests/docker_tests/test_bash_executor.py
```diff
@@ -146,7 +170,7 @@ async def test_unlimited_output_and_decoding(self):
             max_output_bytes=None,
         )
 
-        assert stdout == "  indented�"
+        assert stdout == "  indented�\n"
```
samorojy: ?

samorojy (approved):

PR #290, opened 2026-08-31
[JBRes-10676] Binary-safe chunked file transfer (stack 2/8)

vpoliakov-pixel:
**Stack 2/8** — #289 ← **#290** ← #291 ← #292 ← #293 ← #294 ← #295 ← #296 (merge in order, oldest first).
Base: `…-01-…`. Branch: `vladimir.poliakov/jbres-10676-02-…`.

Improvement #6 of JBRes-10676. Sits above the bash-logging fix because that is what stops a large transfer from flooding the log; otherwise independent.

## What changed

Moving a binary file in or out of a sandbox meant base64 through the bash tool, chunked both ways: bounded on the way in by the 128 KiB ceiling on a single `execve` argument, and logged twice per chunk on the way out.

Two endpoints carry the bytes instead. `POST /api/tools/file/upload` writes one base64 chunk at an offset; `POST /api/tools/file/download` reads one back. They ride the **existing JSON forwarding path**, which is precisely what makes them binary-safe — the orchestrator stores and replays a forwarded request as JSON *text*, so raw bytes would not survive the round trip but base64 does.

On the client, `server.upload_file` / `upload_bytes` / `download_file` / `download_bytes` do the chunking, streaming through a thread so a large file is bounded by the sandbox's disk rather than by client memory.

```mermaid
flowchart LR
  c["client.upload_file (new)"] --> f["POST /api/forward/... (unchanged)"]
  f --> u["POST /api/tools/file/upload (new)"]
  u --> w["FileManager.write_chunk (new)"] --> disk[("file in sandbox")]
  disk --> r["FileManager.read_chunk (new)"] --> d["POST /api/tools/file/download (new)"]
  d --> f2["forwarding, JSON text (unchanged)"] --> c2["client.download_file (new)"]
  old["base64 via bash tool (was: only option)"] -.->|replaced| u
```

## How to verify

```
uv run pytest -m unit unit-tests/test_file_transfer.py
```

- `unit-tests/test_file_transfer.py` — chunk primitives, both tool-service actions, and the client loops against an in-memory sandbox. Every payload is deliberately non-UTF-8, so any hop that decoded it fails the comparison.
- `e2e-tests/test_file_transfer.py` — the same round trip through a real orchestrator, which is the hop the feature exists for. **Not run locally** (no cluster); collected only.

Full suite on this branch: **1192 passed, 2 skipped**; docs site builds.

Resolves: JBRes-10676 (partially — #6)

### DIFF
--- a/client/src/idegym/client/operations/files.py
+++ b/client/src/idegym/client/operations/files.py
@@ -248,10 +248,11 @@ class FileOperations:
         offset = 0
         # An empty source still sends one request, otherwise the file would never be created.
         while True:
+            chunk = await read_chunk_at(offset)
             response = await self.upload_chunk(
                 server_id=server_id,
                 file_path=file_path,
-                data=await read_chunk_at(offset),
+                data=chunk,
                 offset=offset,
                 client_id=client_id,
                 request_timeout=request_timeout,
@@ -260,6 +261,14 @@ class FileOperations:
             offset += response.bytes_written
             if offset >= total:
                 return response.size
+            if not chunk:
+                # The loop advances by bytes written, so a short read before `total` would spin
+                # forever sending empty chunks. Happens when the source shrinks mid-upload:
+                # `total` was measured up front, and reads past the new end return nothing.
+                raise RuntimeError(
+                    f"Upload source ended at {offset} bytes, before the expected {total}: "
+                    f"it shrank while {file_path} was being uploaded"
+                )
 
     async def _download_chunks(
         self,
--- a/client/src/idegym/client/operations/files.py
+++ b/client/src/idegym/client/operations/files.py
@@ -1,8 +1,22 @@
-from typing import Optional
+import asyncio
+from base64 import b64decode, b64encode
+from collections.abc import AsyncIterator, Awaitable, Callable
+from pathlib import Path
+from typing import BinaryIO, Optional
 from uuid import UUID
 
 from idegym.api.paths import ToolsPath
-from idegym.api.tools.file import CreateFileRequest, EditFileRequest, FileResult, PatchFileRequest
+from idegym.api.tools.file import (
+    DEFAULT_FILE_CHUNK_BYTES,
+    CreateFileRequest,
+    DownloadFileChunkRequest,
+    DownloadFileChunkResponse,
+    EditFileRequest,
+    FileResult,
+    PatchFileRequest,
+    UploadFileChunkRequest,
+    UploadFileChunkResponse,
+)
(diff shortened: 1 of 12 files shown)


vpoliakov-pixel:
## Manual verification — binary round trip through a real orchestrator (needs a cluster)

The unit tests prove the chunking and the encoding against an in-memory sandbox, but the claim
this PR actually makes is about *the orchestrator hop*: it stores and replays a forwarded
request as JSON text. Only a cluster exercises that. Not run here — no minikube on this
machine.

```sh
uv run pytest e2e-tests/test_file_transfer.py -m e2e -q
```

**Expect:** 3 passed. The payload is `bytes(range(256)) * 64` — 16 KiB containing every byte
value, including a lone `0xFF` that is not valid UTF-8 — uploaded in 4 KiB chunks and compared
byte-for-byte on the way back, both into memory and onto disk.

**Failure looks like:**
- `AssertionError` on the byte comparison with `�` (U+FFFD) at the position of every
  non-ASCII byte → some hop decoded the payload as text instead of passing the base64 through.
- A short read where `len(result) < len(BINARY)` → the `eof`/offset loop is terminating early.
- `wc -c` reporting the wrong size in `test_uploaded_file_is_visible_to_the_bash_tool` → the
  chunk offsets or the `truncate` semantics are wrong, not the encoding.

### Spot check by hand, if you would rather not run the suite

```python
payload = bytes(range(256)) * 64
size = await server.upload_bytes("/tmp/blob.bin", payload, chunk_bytes=4096)
assert size == len(payload)
assert await server.download_bytes("/tmp/blob.bin", chunk_bytes=4096) == payload
```

**Expect:** silent success. Any `AssertionError` here is the feature failing at its one job.

client/src/idegym/client/operations/files.py
```diff
@@ -63,3 +77,231 @@ async def patch_file(
...
+        offset = 0
+        # An empty source still sends one request, otherwise the file would never be created.
+        while True:
+            chunk = await read_chunk_at(offset)
+            response = await self.upload_chunk(
+                server_id=server_id,
+                file_path=file_path,
+                data=chunk,
+                offset=offset,
+                client_id=client_id,
```
samorojy: A short read counts as success (files.py:257). _download_chunks stops without an error when it gets an empty chunk before end-of-file, which happens if the remote file shrinks during the transfer. download_bytes and download_file then return a truncated result as if it were complete. The upload side raises RuntimeError in the same situation. Fix: compare offset with response.size and raise.

samorojy (approved):

PR #291, opened 2026-08-31
[JBRes-10676] SDK: typed HTTP errors and opt-in tracing (stack 3/8)

vpoliakov-pixel:
**Stack 3/8** — #289 ← #290 ← **#291** ← #292 ← #293 ← #294 ← #295 ← #296 (merge in order, oldest first).
Base: `…-02-…`. Branch: `vladimir.poliakov/jbres-10676-03-…`.

Improvements #11 and #15 of JBRes-10676 — the two High-priority SDK items. Grouped because both change what `IdeGYMClient` does by default. #11 is the foundation the later SDK PRs build on (#6 in this stack raises these exceptions), so it sits low.

## What changed

**Typed HTTP errors** (#11, flagged in the ticket as the single highest-value SDK change). Every HTTP failure surfaced as a plain `RuntimeError` whose message embedded the status in one of two shapes, so a caller distinguishing "the sandbox is gone" from "the control plane is busy" from "the command timed out" had to parse that message — and would break the day either string changed.

`IdeGYMHTTPError` now carries `status_code`, `body`, `method` and `url`, with subclasses for the distinctions retry policy turns on: `IdeGYMNotFoundError` (404 **and** 410 — a pod the orchestrator cannot reach is gone from the caller's side), `IdeGYMBusyError` (429, 503), `IdeGYMTimeoutError` (408, 504, and a client-side httpx timeout), plus Auth, BadRequest, Cancelled and ServerError.

**Tracing is opt-in** (#15). Constructing a client without an explicit `OTELConfig` fell back to a hardcoded remote collector, exporting telemetry off the user's infrastructure silently. The endpoint now comes only from `IDEGYM_OTEL_TRACING_ENDPOINT` or an explicit config.

## Backwards compatibility

Deliberately non-breaking: the exceptions also subclass `RuntimeError` and the message text is byte-identical, so existing `except RuntimeError` keeps working. There is a test for exactly that.

## How to verify

```
uv run pytest -m unit unit-tests/test_client_exceptions.py unit-tests/test_client_tracing.py
```

- `test_client_exceptions.py` — every mapped status plus both fallbacks; two raise sites driven through a real `httpx.MockTransport`; the compatibility properties above.
- `test_client_tracing.py` — default-off, both ways to turn it on, and a blank env var (pydantic rejects it as a URL, so it must read as "unset").

Full suite on this branch: **1218 passed, 2 skipped**; docs site builds.

Resolves: JBRes-10676 (partially — #11, #15)

### DIFF
--- a/client/src/idegym/client/client.py
+++ b/client/src/idegym/client/client.py
@@ -37,6 +37,7 @@ from idegym.api.pod_spec import (
 )
 from idegym.api.resources import KubernetesResources
 from idegym.api.type import KubernetesNodeSelector, KubernetesObjectName, OCIImageName
+from idegym.client.exceptions import http_error
 from idegym.client.operations.clients import ClientOperations
 from idegym.client.operations.forwarding import ForwardingOperations
 from idegym.client.operations.jobs import JobOperations
@@ -362,7 +363,8 @@ class IdeGYMClient:
         """
         Start an IdeGYM server and return an :class:`IdeGYMServer` handle.
 
-        Raises ``RuntimeError`` if the orchestrator returns an error response.
+        Raises an :class:`~idegym.client.exceptions.IdeGYMHTTPError` subclass if the orchestrator
+        returns an error response.
         Prefer :meth:`with_server` for automatic cleanup.
         """
         logger.info(f"Starting IdeGYM server: name={server_name}, image={image_tag}")
@@ -392,8 +394,11 @@ class IdeGYMClient:
         )
 
         if isinstance(server_response, ErrorResponse):
-            # Branching on a response union, not validating an argument's type — RuntimeError is correct.
-            raise RuntimeError(f"Failed to start server: {server_response.model_dump()}")  # noqa: TRY004
+            raise http_error(
+                f"Failed to start server: {server_response.model_dump()}",
+                status_code=server_response.status_code,
+                body=server_response.body,
+            )
         elif isinstance(server_response, StartServerResponse) and server_response.server_id:
             return IdeGYMServer(
                 server_id=server_response.server_id,
--- a/client/src/idegym/client/client.py
+++ b/client/src/idegym/client/client.py
@@ -91,7 +91,8 @@ class IdeGYMClient:
             heartbeat_interval_in_seconds: Interval between heartbeat requests.
             request_timeout_in_seconds: Default timeout for every HTTP request.
             otel_config: OpenTelemetry configuration for tracing. Falls back to ``IDEGYM_OTEL_*``
-                environment variables when not provided.
+                environment variables when not provided. Tracing stays off unless an endpoint is
+                configured, either here or through ``IDEGYM_OTEL_TRACING_ENDPOINT``.
         """
         if orchestrator_url == "idegym.test":
             orchestrator_url = f"http://{orchestrator_url}"
@@ -120,10 +121,12 @@ class IdeGYMClient:
             ),
         )
 
+        # Tracing is opt-in: with no endpoint the exporter is never built, so a caller who
+        # does not know about OTEL cannot end up shipping telemetry off their infrastructure.
         otel_config = otel_config or OTELConfig(
             service_name=env.get("IDEGYM_OTEL_SERVICE_NAME", generate_service_name()),
             tracing=TracingConfig(
-                endpoint=env.get("IDEGYM_OTEL_TRACING_ENDPOINT", "https://tempo.labs.jb.gg/v1/traces"),
+                endpoint=env.get("IDEGYM_OTEL_TRACING_ENDPOINT", "").strip() or None,
                 timeout=int(env.get("IDEGYM_OTEL_TRACING_TIMEOUT", "10")),
                 auth=BasicAuth(
                     username=env.get("IDEGYM_OTEL_TRACING_AUTH_USERNAME"),
--- a/client/src/idegym/client/exceptions.py
+++ b/client/src/idegym/client/exceptions.py
@@ -1,0 +1,116 @@
+"""Typed failures for IdeGYM SDK calls.
+
+A retry policy has to tell "the sandbox is gone" from "the control plane is busy" from "the
+command timed out". Every failure used to arrive as a plain ``RuntimeError`` whose message
+embedded the status, which left callers parsing message text that changes whenever the format
+does. These exceptions carry the status code and the response body as attributes instead.
+
+They subclass ``RuntimeError`` as well as ``IdeGYMException`` so that code written against the
+old behaviour — including ``except RuntimeError`` around a client call — keeps working, and the
+messages are unchanged for the same reason.
+"""
+
+from http import HTTPStatus
+from typing import Optional
+
+from idegym.api.exceptions import IdeGYMException
+
+
+class IdeGYMHTTPError(IdeGYMException, RuntimeError):
+    """A call to the orchestrator, or to a server through it, failed.
+
+    ``status_code`` is the HTTP status the failure carried. It is ``None`` when the request
+    never produced one — a client-side timeout, for instance.
+    """
+
+    def __init__(
+        self,
+        message: str,
+        *,
+        status_code: Optional[int] = None,
+        body: Optional[str] = None,
+        method: Optional[str] = None,
+        url: Optional[str] = None,
+    ) -> None:
+        super().__init__(message)
+        self.message = message
+        self.status_code = status_code
+        self.body = body
+        self.method = method
+        self.url = url
+
+
+class IdeGYMBadRequestError(IdeGYMHTTPError):
+    """The request was rejected as malformed or invalid. Retrying it unchanged will not help."""
+
+
+class IdeGYMAuthError(IdeGYMHTTPError):
+    """Credentials were missing, wrong, or insufficient for the resource."""
+
+
+class IdeGYMNotFoundError(IdeGYMHTTPError):
+    """The addressed client, server, or operation does not exist any more.
+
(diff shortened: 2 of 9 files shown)


client/src/idegym/client/operations/forwarding.py
```diff
@@ -56,5 +57,12 @@ async def forward_request(
             if response.status_code == HTTPStatus.TOO_EARLY.value:
                 raise InspectionsNotReadyException()
 
-            # Branching on a response union, not validating an argument's type — RuntimeError is correct.
-            raise RuntimeError(f"Failed to forward request {method} {url}: {response.model_dump()}")  # noqa: TRY004
+            # The message is deliberately unchanged: it predates the typed exceptions and
+            # callers parse it. New code should read `status_code` and `body` instead.
+            raise http_error(
```
samorojy: Is a 404/503 returned *by the env-server inside the sandbox* distinguishable from the orchestrator's own 404/410 here? As I read it, the forwarded status goes straight through `http_error`, so an app-level `404 Path not found` from a live sandbox (e.g. `server/router/fs.py`) becomes `IdeGYMNotFoundError`, and `client.md` tells users to treat that as "the sandbox is gone — start a new one".

Scenario: agent reads a missing file via `forward_request` → sandbox returns 404 → caller following the docs tears down a healthy sandbox and loses the episode.

Suggestion: use a distinct type (or a flag like `from_sandbox=True`) for relayed sandbox responses, or keep `NotFound` for orchestrator-level 404/410 only.

client/src/idegym/client/operations/utils.py
```diff
@@ -122,15 +123,28 @@ async def make_request(
             logger.warning(f"Request cancelled: url={url}")
             raise
 
+        except TimeoutException as ex:
+            message = f"Request timed out: url={url} error='{ex}'"
+            logger.error(message)
+            raise IdeGYMTimeoutError(message, method=method, url=url)
```
samorojy: Two things on this raise:

1. **Compatibility:** `IdeGYMTimeoutError` isn't an `httpx` exception, so callers that currently do `except httpx.TimeoutException` / `except httpx.HTTPError` around client calls will stop catching timeouts. That goes against the "existing handlers keep working" promise. At minimum this deserves a changelog note.
2. **Chaining:** please add `from ex` here, and on the `HTTPStatusError` path below. Without it `__cause__` is `None` and the traceback reads "During handling of the above exception, another exception occurred", so it's harder to tell a `ConnectTimeout` from a `ReadTimeout` or a `PoolTimeout`.

client/src/idegym/client/operations/utils.py
```diff
@@ -122,15 +123,28 @@ async def make_request(
...
             )
             logger.error(message)
-            raise RuntimeError(message)
+            raise http_error(
+                message,
+                status_code=ex.response.status_code,
+                body=ex.response.text,
+                method=method,
+                url=url,
+            )
```
samorojy: Non-timeout transport failures (`httpx.ConnectError`, `RemoteProtocolError`, `ReadError`) still fall through to the bare `except Exception: raise` ([L153](https://github.com/JetBrains-Research/idegym/blob/3ed45b7c6a94a3060862ed73abcab35fafa9b9a1/client/src/idegym/client/operations/utils.py#L153-L155)) and escape as raw httpx errors. But `client.md` says "An HTTP failure raises a subclass of `IdeGYMHTTPError`".

Scenario: the orchestrator pod restarts during a rollout → `ConnectError` → a retry policy that branches on `IdeGYMHTTPError`/`IdeGYMBusyError` misses the most retryable failure of all.

Suggestion: wrap `httpx.TransportError` in a typed error (with `status_code=None`), or narrow the docs.

samorojy (commented):
A few concerns about how the new error taxonomy lines up with what callers actually see. Details are inline.

One more point is outside the diff, so it goes here: `stop_server`/`finish_server` use `@retry_with_backoff(attempts=3)` ([client.py#L316](https://github.com/JetBrains-Research/idegym/blob/3ed45b7c6a94a3060862ed73abcab35fafa9b9a1/client/src/idegym/client/client.py#L316)), which retries on any `Exception`. Now that errors are typed, it would be good to skip retries for non-retryable `IdeGYMNotFoundError` / `IdeGYMAuthError` / `IdeGYMBadRequestError`. A server that was already reaped during `with_server` cleanup currently gets 3 attempts and 3 logged exceptions before the `NotFound` is re-raised.

_These come from reading the code, not from running it. Tell me if I've misread something._

website/docs/reference/client.md
```diff
@@ -485,27 +485,73 @@ env_client = MyOpenEnvClient(base_url=url)
...
+except IdeGYMHTTPError as e:
+    print(f"Failed with {e.status_code}: {e.body}")
 ```
 
+| Exception | Statuses | What it means for a retry |
+|-----------|----------|---------------------------|
+| `IdeGYMBadRequestError` | 400, 422, other 4xx | The request is wrong; retrying it unchanged will not help |
+| `IdeGYMAuthError` | 401, 403 | Credentials are missing, wrong, or insufficient |
+| `IdeGYMNotFoundError` | 404, 410 | The client, server, or operation is gone — including a pod the orchestrator can no longer reach |
+| `IdeGYMTimeoutError` | 408, 504, client-side timeout | Safe to retry if the operation is idempotent |
```
samorojy: The timeouts users actually hit on the client side still raise the builtin `TimeoutError`, not `IdeGYMTimeoutError`:
- `asyncio.timeout(...)` around async-op polling: [utils.py#L177](https://github.com/JetBrains-Research/idegym/blob/3ed45b7c6a94a3060862ed73abcab35fafa9b9a1/client/src/idegym/client/operations/utils.py#L177)
- `start_server`'s explicit `raise TimeoutError(...)`: [servers.py#L84](https://github.com/JetBrains-Research/idegym/blob/3ed45b7c6a94a3060862ed73abcab35fafa9b9a1/client/src/idegym/client/operations/servers.py#L84), [#L150](https://github.com/JetBrains-Research/idegym/blob/3ed45b7c6a94a3060862ed73abcab35fafa9b9a1/client/src/idegym/client/operations/servers.py#L150-L152)

So a long `tools/bash` call that exceeds `wait_timeout_in_sec` escapes an `except IdeGYMTimeoutError` handler, and an `except RuntimeError` one too. Either convert these, or say in this row that only HTTP-transport timeouts are covered.

website/docs/reference/http_error_codes.md
```diff
@@ -5,6 +5,19 @@ the **Orchestrator** (Kubernetes orchestration service) and the **IdeGYM Server*
...
 ---
 
+## Reading these codes from the Python client
+
+The client turns each of these into a typed exception rather than a bare `RuntimeError`:
+`IdeGYMNotFoundError` for 404 and 410, `IdeGYMBusyError` for 429 and 503, `IdeGYMTimeoutError`
+for 408 and 504, and so on. The status and body are attributes on the exception. The full
+mapping is in [Client Library — Error Handling](client.md#error-handling).
+
+A failure reported through an async operation's `result.status_code` is mapped the same way, so
```
samorojy: I don't think this holds for start-server. `ServerOperations.start_server` swallows a 429 `ErrorResponse` and retries ([servers.py#L141-L154](https://github.com/JetBrains-Research/idegym/blob/3ed45b7c6a94a3060862ed73abcab35fafa9b9a1/client/src/idegym/client/operations/servers.py#L141-L154)). When the budget runs out it raises the builtin `TimeoutError("Server start timed out ...")`, never `IdeGYMBusyError`, so `except IdeGYMBusyError: back off` won't fire when the quota is exhausted.

Relatedly, a failed client registration in `__aenter__` still raises a plain `RuntimeError` with no `status_code` ([client.py#L208](https://github.com/JetBrains-Research/idegym/blob/3ed45b7c6a94a3060862ed73abcab35fafa9b9a1/client/src/idegym/client/client.py#L208)).

client/src/idegym/client/client.py
```diff
@@ -119,10 +121,12 @@ def __init__(
             ),
         )
 
+        # Tracing is opt-in: with no endpoint the exporter is never built, so a caller who
+        # does not know about OTEL cannot end up shipping telemetry off their infrastructure.
         otel_config = otel_config or OTELConfig(
             service_name=env.get("IDEGYM_OTEL_SERVICE_NAME", generate_service_name()),
             tracing=TracingConfig(
-                endpoint=env.get("IDEGYM_OTEL_TRACING_ENDPOINT", "https://tempo.labs.jb.gg/v1/traces"),
+                endpoint=env.get("IDEGYM_OTEL_TRACING_ENDPOINT", "").strip() or None,
```
samorojy: I understand making tracing opt-in is intentional (and the rationale above makes sense). But anyone who relied on the old default endpoint will silently stop getting spans after upgrading, and the only signal is an info-level "Tracing not enabled" line. Could we add a changelog entry, and maybe a one-time warning when no endpoint is configured?

client/src/idegym/client/exceptions.py
```diff
@@ -0,0 +1,116 @@
...
+        self,
+        message: str,
+        *,
+        status_code: Optional[int] = None,
+        body: Optional[str] = None,
+        method: Optional[str] = None,
+        url: Optional[str] = None,
+    ) -> None:
+        super().__init__(message)
+        self.message = message
```
samorojy: nit: `self.message` duplicates `args[0]` / `str(e)`. Is it needed? Also, `error_class_for_status` and `http_error` are both public entry points for the same lookup. Consider keeping only `http_error` public.

samorojy (approved):

PR #292, opened 2026-08-31
[JBRes-10676] Bash tool: per-command context, script size, init semantics (stack 4/8)

vpoliakov-pixel:
**Stack 4/8** — #289 ← #290 ← #291 ← **#292** ← #293 ← #294 ← #295 ← #296 (merge in order, oldest first).
Base: `…-03-…`. Branch: `vladimir.poliakov/jbres-10676-04-…`.

Improvements #3, #4 and #5 of JBRes-10676 — the three Medium bash-tool items. Grouped because they all rewrite the same execution path and only make sense read together. Depends on #1 in this stack (same files).

## What changed

**`cwd`, `env` and `user` on the request** (#3). The working directory was fixed per server and the environment cleaned, so every caller synthesized context into the script: `cd -- … || exit 1`, one `export` per variable, a `runuser` wrapper. That is error-prone shell each client reimplements, it puts secrets in the command text, and it spends the script's size budget.

**The script runs from a temp file** (#4). It travelled as one argument to `bash -c`, so `MAX_ARG_STRLEN` (128 KiB) was a hard ceiling and an oversized script failed with a bare `E2BIG`. A file, not bash's stdin, on purpose: a script read from stdin is consumed incrementally, so any command in it that reads stdin swallows the remainder.

**The init prefix no longer rewrites the script** (#5). It was `source <init> && <command>`, so only the *first* statement was conditional on the init and the rest ran regardless — `a; b; c` did not mean what the caller wrote. Now joined with `;`, on the same line so bash still reports the caller's own line numbers.

```mermaid
flowchart TD
  req["BashCommandRequest (+ cwd, env, user)"] --> resolve["resolve_working_directory (new)"]
  resolve --> join["'source init ; cmd' (was: '&&')"]
  join --> write["_write_script: temp file (was: bash -c arg, 128 KiB cap)"]
  write --> argv["bash file, or runuser -u user (new)"]
  argv --> spawn["create_subprocess_exec(cwd, env=cleanenv() | env)"]
  spawn --> cleanup["finally: reap + _remove_script (new)"]
```

## How to verify

```
uv run pytest -m unit unit-tests/test_bash_executor.py
uv run pytest integration-tests/docker_tests/test_bash_executor.py integration-tests/test_tool_service.py
```

Integration proves the parts unit tests cannot: a **512 KiB script runs** (4× the old ceiling), a script whose first command reads stdin still runs its second, `a; b; c` all execute while the caller's own `&&` still short-circuits, and an `env` value never reaches the logged command.

An empty script is now a no-op exiting 0 rather than a syntax error — same change seen from the other side.

Full suite: **1232 passed, 2 skipped**; docs site builds.

Resolves: JBRes-10676 (partially — #3, #4, #5)

### DIFF
--- a/backend-utils/src/idegym/backend/utils/bash_executor.py
+++ b/backend-utils/src/idegym/backend/utils/bash_executor.py
@@ -1,8 +1,9 @@
 import asyncio
+import contextlib
 import os
 import re
-import shlex
 import signal
+import tempfile
 from asyncio.subprocess import Process
 from collections import deque
 from importlib.resources import files
@@ -186,16 +187,43 @@ def _command_excerpt(command: str) -> str:
     return _log_excerpt(_redact_exports(command))
 
 
-def _shell_invocation(bash_command: str, user: Optional[str]) -> str:
-    """Build the shell line that runs the script, optionally dropping to another user.
+def _process_argv(script_path: str, user: Optional[str]) -> list[str]:
+    """Build the argv that runs the script file, optionally dropping to another user.
 
     ``runuser`` is used rather than ``su`` because it does not authenticate and keeps the
     caller's environment, which is what the ``env`` argument has already been merged into.
     """
-    invocation = f"bash -c {shlex.quote(bash_command)}"
+    invocation = ["bash", script_path]
     if user is None:
         return invocation
-    return f"runuser --preserve-environment -u {shlex.quote(user)} -- {invocation}"
+    return ["runuser", "--preserve-environment", "-u", user, "--", *invocation]
+
+
+def _write_script(script: str, readable_by_other_user: bool) -> str:
+    """Write the script to a temp file and return its path.
+
+    Passing the script as a ``bash -c`` argument capped it at Linux's ``MAX_ARG_STRLEN``
+    (128 KiB), and an oversized script failed with a bare ``E2BIG`` rather than anything a
+    caller could act on. A file has no such ceiling, and unlike feeding bash on stdin it leaves
+    the command's own stdin alone — a script read from stdin is consumed incrementally, so any
+    command inside it that reads stdin would swallow the rest of the script.
+    """
+    descriptor, path = tempfile.mkstemp(prefix="idegym-bash-", suffix=".sh")
+    try:
+        with os.fdopen(descriptor, "w", encoding="utf-8") as handle:
+            handle.write(script)
+        if readable_by_other_user:
+            # mkstemp creates 0600, which the target user of `runuser` could not read.
+            os.chmod(path, 0o644)
+    except BaseException:
+        _remove_script(path)
+        raise
+    return path
+
+
+def _remove_script(path: str) -> None:
+    with contextlib.suppress(OSError):
+        os.unlink(path)
 
 
 def _signal_process_group(process: Process, requested_signal: signal.Signals) -> bool:
@@ -295,7 +323,8 @@ class BashExecutor:
         to run as root, since it shells out through ``runuser``.
 
         Output is returned verbatim unless ``strip_output`` asks for surrounding
-        whitespace to be trimmed.
+        whitespace to be trimmed. The script itself is written to a temp file and run as
+        ``bash <file>``, so its size is not capped by the kernel's argument limit.
 
         Returns a tuple of (stdout, stderr, exit_code).
         Raises BashCommandExecutionTimeoutError if the timeout is exceeded.
@@ -313,14 +342,19 @@ class BashExecutor:
         logger.debug("Bash command", command=_command_excerpt(command))
 
         bash_command = f"source {__BASH_INIT_FILEPATH__} && {command}"
-        process = await asyncio.create_subprocess_shell(
-            cmd=_shell_invocation(bash_command, user),
-            stdout=asyncio.subprocess.PIPE,
-            stderr=asyncio.subprocess.PIPE,
-            cwd=working_directory,
-            preexec_fn=os.setsid,
-            env=cleanenv() | (env or {}),
-        )
+        script_path = await asyncio.to_thread(_write_script, bash_command, user is not None)
+        try:
+            process = await asyncio.create_subprocess_exec(
+                *_process_argv(script_path, user),
+                stdout=asyncio.subprocess.PIPE,
+                stderr=asyncio.subprocess.PIPE,
+                cwd=working_directory,
+                preexec_fn=os.setsid,
+                env=cleanenv() | (env or {}),
+            )
+        except BaseException:
+            _remove_script(script_path)
+            raise
 
         communication_task = asyncio.create_task(
             _communicate_bounded(process, stdout_collector, stderr_collector),
@@ -345,6 +379,7 @@ class BashExecutor:
             await _finish_output_drain(process, communication_task)
             _close_output_pipes(process)
             await _reap_process(process)
+            _remove_script(script_path)
 
         if timed_out:
             logger.warning(
--- a/backend-utils/src/idegym/backend/utils/bash_executor.py
+++ b/backend-utils/src/idegym/backend/utils/bash_executor.py
@@ -186,6 +186,18 @@ def _command_excerpt(command: str) -> str:
     return _log_excerpt(_redact_exports(command))
 
 
+def _shell_invocation(bash_command: str, user: Optional[str]) -> str:
+    """Build the shell line that runs the script, optionally dropping to another user.
+
+    ``runuser`` is used rather than ``su`` because it does not authenticate and keeps the
+    caller's environment, which is what the ``env`` argument has already been merged into.
+    """
+    invocation = f"bash -c {shlex.quote(bash_command)}"
+    if user is None:
+        return invocation
+    return f"runuser --preserve-environment -u {shlex.quote(user)} -- {invocation}"
+
+
 def _signal_process_group(process: Process, requested_signal: signal.Signals) -> bool:
     """Signal a process group, tolerating races after its leader has exited."""
     try:
@@ -245,6 +257,19 @@ class BashExecutor:
     def __init__(self, working_directory: Optional[Path] = None):
         self.working_directory = working_directory
 
+    def resolve_working_directory(self, cwd: Optional[str]) -> Optional[Path]:
+        """Resolve a per-command ``cwd`` against the executor's directory.
+
+        An absolute path is used as given; a relative one is taken from the executor's working
+        directory, which is the server's project root.
+        """
+        if cwd is None:
+            return self.working_directory
+        requested = Path(cwd)
+        if requested.is_absolute() or self.working_directory is None:
+            return requested
+        return self.working_directory / requested
+
     async def execute_bash_command(
         self,
         command: str,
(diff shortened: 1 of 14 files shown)


backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -186,6 +197,82 @@ def _command_excerpt(command: str) -> str:
...
+def _process_argv(script_path: str, user: Optional[str]) -> list[str]:
+    """Build the argv that runs the script file, optionally dropping to another user.
+
+    ``runuser`` is used rather than ``su`` because it does not authenticate and keeps the
+    caller's environment, which is what the ``env`` argument has already been merged into.
+    """
+    invocation = ["bash", script_path]
+    if user is None:
+        return invocation
+    return ["runuser", "--preserve-environment", "-u", user, "--", *invocation]
```
samorojy: **`user=` fails in the default deployment.** `runuser` can only be run by root, but the project's server image runs as `USER appuser` (see `image-builder/src/idegym/image/templates/server.Dockerfile.jinja`). So `server.execute_bash('id', user='appuser')` returns exit_code=1 with `runuser: may not be used by non-root users`, reported as a normal command failure and not a 400. appuser has passwordless sudo, so `sudo -u` would work. Otherwise, check `os.geteuid() == 0` up front, and skip the switch when `user` is already the current user.

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -186,6 +197,82 @@ def _command_excerpt(command: str) -> str:
...
+    statement was conditional on the init succeeding and the rest ran regardless, so
+    ``a; b; c`` did not mean what the caller wrote — clients defended by wrapping every script
+    in a brace group. Keeping the prefix on the same line also keeps the caller's line numbers
+    intact, so a bash error still points at the right line of their script.
+    """
+    init = shlex.quote(str(__BASH_INIT_FILEPATH__))
+    # `;` rather than `&&` so the caller's script keeps its own semantics, but the init is still
+    # guarded: without this a missing or unreadable init left the script running in an
+    # unconfigured shell and failing later as "command not found", with the caller's exit code.
+    guard = f'source {init} || {{ echo "IdeGYM: failed to source the bash integration at {init}" >&2 ; exit 1 ; }}'
```
samorojy: **The init guard misfires on a normal `~/.bashrc`.** `source init || { ...; exit 1; }` checks the status of the last command in the sourced file. If an image's `~/.bashrc` ends with `[ -f ~/.fzf.bash ] && source ~/.fzf.bash` and that file is absent, `source` returns 1. Every command then fails with `IdeGYM: failed to source the bash integration`, and none of the caller's statements run (before this PR only the first statement was skipped). Suggest checking readability separately: `[ -r init ] || {...}; source init`, or ignore the status of the source.

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -186,6 +197,82 @@ def _command_excerpt(command: str) -> str:
...
+    return {"HOME": entry.pw_dir, "USER": user, "LOGNAME": user, "SHELL": entry.pw_shell or "/bin/bash"}
+
+
+def _process_argv(script_path: str, user: Optional[str]) -> list[str]:
+    """Build the argv that runs the script file, optionally dropping to another user.
+
+    ``runuser`` is used rather than ``su`` because it does not authenticate and keeps the
+    caller's environment, which is what the ``env`` argument has already been merged into.
+    """
+    invocation = ["bash", script_path]
```
samorojy: **Bare `bash`/`runuser` are looked up in the caller's PATH.** With `env=environment`, subprocess resolves the executable against the PATH in that env. `execute_bash('mytool', env={'PATH': '/opt/tool/bin'})` (the dict can't expand `$PATH`, so callers pass literal values) makes `create_subprocess_exec('bash', ...)` raise FileNotFoundError, which the global handler returns as a confusing 404. Use `/bin/bash` or resolve the path with the server's own PATH.

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -186,6 +197,82 @@ def _command_excerpt(command: str) -> str:
...
+    try:
+        with os.fdopen(descriptor, "w", encoding="utf-8") as handle:
+            handle.write(script)
+        if readable_by_other_user:
+            # mkstemp creates 0600, which the target user of `runuser` could not read.
+            # TODO: 0644 also exposes the script to any co-tenant process for the duration of the
+            # run, which matters because callers do put credentials in scripts (see
+            # `_redact_exports`). Prefer `os.chown(path, uid, -1)` with 0600, or a directory only
+            # the target user can traverse.
+            os.chmod(path, 0o644)
```
samorojy: **The script is world-readable while it runs.** With `user` set, the whole script, inline credentials included, sits at 0644 in the shared temp dir for the entire run. For example, `execute_bash("export TOKEN=abc; ./deploy", user='dev')` lets anyone in the container `cat /tmp/idegym-bash-*.sh`. `_redact_exports` was added because callers put secrets in scripts, so this weakens that. The TODO notes the issue, but it would be good to fix it before merge, e.g. `chown` the file to the target user with 0600, or pass the script over stdin.

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -261,26 +370,46 @@ async def execute_bash_command(
...
 
         Returns a tuple of (stdout, stderr, exit_code).
         Raises BashCommandExecutionTimeoutError if the timeout is exceeded.
         """
         stdout_collector = _OutputCollector(max_output_bytes)
         stderr_collector = _OutputCollector(max_output_bytes)
-        logger.info("Executing bash command", command_chars=len(command))
+        working_directory = self.resolve_working_directory(cwd)
+        # The user's identity goes on before the caller's env, so an explicit HOME still wins.
+        environment = cleanenv() | (_user_environment(user) if user is not None else {}) | (env or {})
```
samorojy: **`env` isn't validated, so bad input returns a 500.** `env={'A=B': '1'}`, `env={'': 'x'}` or `env={'K': 'a\x00b'}` makes `create_subprocess_exec` raise ValueError, which becomes a 500 with a traceback. Bad `cwd`/`user` values are mapped to 400 carefully. A pydantic validator on `BashCommandRequest.env` would give a 400 here too.

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -261,26 +370,46 @@ async def execute_bash_command(
...
-        process = await asyncio.create_subprocess_shell(
-            cmd=f"bash -c {shlex.quote(bash_command)}",
-            stdout=asyncio.subprocess.PIPE,
-            stderr=asyncio.subprocess.PIPE,
-            cwd=self.working_directory,
-            preexec_fn=os.setsid,
-            env=cleanenv(),
-        )
+        bash_command = _prepend_bash_integration(command)
+        script_path = await asyncio.to_thread(_write_script, bash_command, user is not None)
```
samorojy: **The temp file leaks if the request is cancelled.** If the request is cancelled while `asyncio.to_thread(_write_script, ...)` is running, the worker thread still finishes `mkstemp`, but the path is never returned, so the file is never removed (and may be 0644 with secrets in it). Consider creating the file name or fd on the event-loop side before handing off, or cleaning up in a finally block that can reach the path.

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -34,6 +37,14 @@ class BashCommandExecutionTimeoutError(BashExecutorError):
     pass
 
 
+class BashExecutorUnknownUserError(BashExecutorError):
```
samorojy: **The new exceptions bypass the project's exception convention.** `BashExecutorUnknownUserError` and `BashExecutorWorkingDirectoryError` extend `BashExecutorError(Exception)`, not `IdeGYMException`, and are turned into 400s by a one-off try/except in a single router. AGENTS.md §3.7 asks for `IdeGYMException` subclasses translated by the global handlers. As written, the same errors from other `BashExecutor` callers (rewards, project reset) return 500s, the 400 body is FastAPI's `{detail}` and not the documented `{timestamp,message,traceback}`, and `http_error_codes.md` still says bash only returns 500 on timeout.

backend-utils/src/idegym/backend/utils/bash_executor.py
```diff
@@ -186,6 +197,82 @@ def _command_excerpt(command: str) -> str:
...
+def _write_script(script: str, readable_by_other_user: bool) -> str:
+    """Write the script to a temp file and return its path.
+
+    Passing the script as a ``bash -c`` argument capped it at Linux's ``MAX_ARG_STRLEN``
+    (128 KiB), and an oversized script failed with a bare ``E2BIG`` rather than anything a
+    caller could act on. A file has no such ceiling, and unlike feeding bash on stdin it leaves
+    the command's own stdin alone — a script read from stdin is consumed incrementally, so any
+    command inside it that reads stdin would swallow the rest of the script.
+    """
+    descriptor, path = tempfile.mkstemp(prefix="idegym-bash-", suffix=".sh")
```
samorojy: **Running `bash <tempfile>` changes `$0` and stderr.** Previously `execute_bash('nosuchcmd')` printed `bash: line 1: nosuchcmd: command not found`. It now prints `/tmp/idegym-bash-k3j9x.sh: line 1: ...` with a different name on each call, which breaks callers or reward checks that compare stderr, and `$(dirname "$0")` now resolves to /tmp. `bash -s` with the script on stdin, or `exec -a bash`, would keep the old behaviour.

website/docs/reference/tools.md
```diff
@@ -47,6 +50,48 @@ is outside the configured limit. Set `max_output_bytes` to `null` only when comp
...
+incrementally, so any command inside it that reads stdin — `cat`, `read`, an interactive
+installer — would swallow the rest of the script. Running from a file leaves the command's own
+stdin alone.
+
+The temp file is removed once the command finishes, including when it times out.
+
+Before the script, IdeGYM sources a bundled init file that sets up the shell environment. It is
+joined to your script with `;`, not `&&`, so it cannot change what your script means: every
+statement runs exactly as written, and your own `&&` and `||` behave normally. The prefix stays
+on the same line, so a bash error still reports the line number you wrote. An empty script is a
```
samorojy: **The docs don't match the endpoint.** This says an empty script is a no-op that exits 0, but `ToolService.execute_tool` (tools/src/idegym/tools/tool_service.py) still rejects an empty command with `ValueError("Missing 'command' ...")`. So `POST /api/tools/bash` with `{"command": ""}` returns a 500. Either drop that check or say this only applies to direct `BashExecutor` calls.

samorojy (approved):
Approving. The per-command context design looks good. Inline comments cover a few issues worth fixing, either here or in a follow-up in the stack. The most important ones: `user=` can't work under the default non-root `appuser` image (it uses `runuser`); the init guard aborts scripts when `~/.bashrc` ends with a non-zero command; and the temp script is world-readable while it runs.

One item outside this diff: `orchestrator/src/idegym/orchestrator/mcp.py` (`run_bash_command`, ~L194) builds `BashCommandRequest` without the new `cwd`/`env`/`user`/`strip_output` fields, so MCP agents can't use per-command context. In particular, they can't pass secrets through `env` and have to fall back to inline `export`s. The frozen payload test in `unit-tests/test_orchestrator_mcp.py` was updated with nulls, but the fields aren't wired through.

PR #293, opened 2026-08-31
[JBRes-10676] Server lifecycle: status, keepalive, listing, reuse reporting (stack 5/8)

vpoliakov-pixel:
**Stack 5/8** — #289 ← #290 ← #291 ← #292 ← **#293** ← #294 ← #295 ← #296 (merge in order, oldest first).
Base: `…-04-bash-execution`. Branch: `vladimir.poliakov/jbres-10676-05-…`.

Improvements #7, #8, #9, #10 of JBRes-10676 — knowing and controlling a server's lifecycle. Grouped because they touch the same four files. **Contains migration 004.**

## What changed

**`GET /api/idegym-servers/{id}/status`** (#7) — there was no "is this server healthy" call, so callers used `list_capabilities` as a probe. It uses an ownership-only lookup rather than `validate_server`, since a status endpoint must *report* a crashed server instead of 410-ing on it, and it records no activity.

**An explicit keepalive** (#8) — the reaper works off `last_heartbeat_time`, which only advances when a request finishes, so a sandbox that is legitimately idle (an agent thinking, a long build) ages toward being reaped while in use. Held in a new column rather than by pushing the heartbeat into the future, so "when was this last used" stays truthful.

**`client.list_servers()`** (#9) — the SDK could not ask what it owned, so finding a leaked pod meant `kubectl`.

**`StartServerResponse.reused`** (#10) — plus the seven-field reuse match key written down. A client that always calls `stop_server` leaves servers `STOPPED`, so reuse can never hit; that was invisible.

```mermaid
flowchart LR
  ka["server.keepalive(minutes) (new)"] --> ep["POST .../keepalive (new)"]
  ep --> col[("servers.keepalive_until (new, migration 004)")]
  hb[("last_heartbeat_time (unchanged)")] --> reaper["watcher cleanup_servers"]
  col --> reaper
  reaper -->|"hold active: skip (new)"| kept["server kept"]
  reaper -->|"no hold, idle past timeout"| kill["clean_up_server, KILLED"]
```

Migration **004** adds nullable `servers.keepalive_until`; chart `schemaRevision` → `004`; single head verified.

## How to verify

```
uv run pytest -m unit unit-tests/test_server_status.py unit-tests/test_server_keepalive.py \
  unit-tests/test_server_listing.py unit-tests/test_server_reuse_reporting.py
```

Keepalive is driven through `cleanup_servers` directly: the behaviour is negative, so a held server must survive a pass that reaps the identical unheld one.

**Not run here (needs Docker):** `integration-tests/database/{test_migration_roundtrip,test_watcher_cleanup}.py`. E2E collected only. Manual recipe in a comment below.

Full suite: **1275 passed, 2 skipped**; docs builds.

Resolves: JBRes-10676 (partially — #7, #8, #9, #10)

### DIFF
--- a/api/src/idegym/api/orchestrator/servers.py
+++ b/api/src/idegym/api/orchestrator/servers.py
@@ -177,6 +177,27 @@ class FinishServerRequest(ServerScopedRequest):
     pass
 
 
+class KeepaliveServerRequest(ServerScopedRequest):
+    """Hold a server against the inactivity reaper for a bounded window."""
+
+    minutes: float = Field(
+        default=15.0,
+        gt=0,
+        le=24 * 60,
+        description=(
+            "How long from now to hold the server, in minutes. Calling again extends the window; "
+            "it is never shortened, so two holders cannot cut each other short."
+        ),
+        examples=[15, 60],
+    )
+
+
+class KeepaliveServerResponse(BaseModel):
+    server_id: int
+    keepalive_until: int = Field(description="Epoch milliseconds until which the server is held", ge=0)
+    minutes: float = Field(description="Minutes from now that 'keepalive_until' corresponds to", ge=0)
+
+
 class RestartServerRequest(ServerScopedRequest):
     server_start_wait_timeout_in_seconds: int = Field(
         default=60, description="Seconds to wait for server readiness after restart", ge=0
@@ -237,6 +258,13 @@ class ServerStatusResponse(BaseModel):
         ge=0,
     )
     idle_seconds: float = Field(description="Seconds since 'last_activity_at'", ge=0)
+    keepalive_until: Optional[int] = Field(
+        default=None,
+        description=(
+            "Epoch milliseconds until which an explicit keepalive holds this server against the "
+            "inactivity reaper, or null when none is in effect"
+        ),
+    )
     pod_phase: Optional[str] = Field(
         default=None, description="Kubernetes phase of the server pod, or null when no pod matches"
     )
--- a/api/src/idegym/api/orchestrator/servers.py
+++ b/api/src/idegym/api/orchestrator/servers.py
@@ -53,7 +53,10 @@ class StartServerRequest(BaseModel):
     )
     server_name: KubernetesObjectName = Field(
         default="default-idegym-server",
-        description="Logical server name used as the Kubernetes resource name prefix and for matching reusable servers",
+        description=(
+            "Logical server name, used as the Kubernetes resource name prefix. It is one of the "
+            "seven fields reuse matches on, not the key: see 'reuse_strategy' for the full set."
+        ),
         examples=["my-server", "echo-env-server"],
     )
     runtime_class_name: Optional[str] = Field(
@@ -135,8 +138,13 @@ class StartServerRequest(BaseModel):
     reuse_strategy: ServerReuseStrategy = Field(
         default=ServerReuseStrategy.RESET,
         description=(
-            "What to do if a server with this name already exists: NONE recreates from scratch, "
-            "RESTART restarts it, RESET resets project state."
+            "Whether to take over an existing server instead of creating one: NONE always creates "
+            "from scratch, RESTART reuses one and restarts its pod, RESET reuses one and resets "
+            "the project state. Reuse runs only for RESTART and RESET, and a candidate must match "
+            "on all seven of: client name, image_tag, runtime_class_name, run_as_root, server_kind, "
+            "server_name (when set), and an availability of FINISHED. A server is FINISHED only "
+            "after finish_server; a client that always calls stop_server leaves its servers STOPPED, "
+            "so reuse never hits. StartServerResponse.reused reports what actually happened."
         ),
     )
     server_kind: ServerKind = Field(
@@ -213,6 +221,14 @@ class StartServerResponse(BaseModel):
     generated_name: Optional[str] = Field(default=None, description="Generated Kubernetes resource name")
     service_name: Optional[str] = Field(default=None, description="Kubernetes Service name for the server")
     image_tag: Optional[str] = Field(default=None)
+    reused: bool = Field(
+        default=False,
+        description=(
+            "True when an existing FINISHED server was taken over rather than a new one created. "
+            "Reuse depends on seven fields matching, so it is easy to configure a request that "
+            "silently never reuses; this reports what happened instead of leaving it to be inferred."
+        ),
+    )
     need_to_reset: bool = Field(default=False, description="True if the reused server requires a project reset")
 
 
--- a/api/src/idegym/api/orchestrator/servers.py
+++ b/api/src/idegym/api/orchestrator/servers.py
@@ -239,6 +239,33 @@ class ServerRequestResponse(BaseModel):
     status: str
 
 
+class ServerSummary(BaseModel):
+    """One row of the server list: what the orchestrator knows without asking Kubernetes.
+
+    Deliberately has no pod fields — listing them would mean one API call per server. Use the
+    status endpoint for the pod view of a single server.
+    """
+
+    server_id: int = Field(description="Numeric IdeGYM server ID")
+    server_name: Optional[str] = Field(default=None, description="Logical server name from the start request")
+    generated_name: str = Field(description="Kubernetes resource name for the server")
+    namespace: str
+    availability: str = Field(description="Availability status recorded by the orchestrator")
+    usable: bool = Field(description="True when the server is in a state that accepts requests")
+    image_tag: Optional[str] = Field(default=None)
+    created_at: int = Field(description="Epoch milliseconds", ge=0)
+    last_activity_at: int = Field(description="Epoch milliseconds", ge=0)
+    keepalive_until: Optional[int] = Field(
+        default=None, description="Epoch milliseconds until which an explicit keepalive holds the server"
+    )
+    details: Optional[str] = Field(default=None, description="Failure reason recorded on a terminal status")
+
+
+class ListServersResponse(BaseModel):
+    client_id: UUID
+    servers: list[ServerSummary] = Field(default_factory=list)
+
+
 class ServerStatusResponse(BaseModel):
     """Everything needed to answer 'is this server usable right now', in one call.
 
--- a/api/src/idegym/api/orchestrator/servers.py
+++ b/api/src/idegym/api/orchestrator/servers.py
@@ -218,6 +218,32 @@ class ServerRequestResponse(BaseModel):
     status: str
 
 
+class ServerStatusResponse(BaseModel):
+    """Everything needed to answer 'is this server usable right now', in one call.
+
+    Reading it does not count as activity, so polling it cannot keep a server alive by accident.
+    """
+
+    server_id: int = Field(description="Numeric IdeGYM server ID")
+    server_name: Optional[str] = Field(default=None, description="Logical server name from the start request")
+    generated_name: str = Field(description="Kubernetes resource name for the server")
+    namespace: str
+    availability: str = Field(description="Availability status recorded by the orchestrator, e.g. ALIVE or CRASHED")
+    usable: bool = Field(description="True when the server is in a state that accepts requests (ALIVE or REUSED)")
+    image_tag: Optional[str] = Field(default=None)
+    created_at: int = Field(description="Epoch milliseconds", ge=0)
+    last_activity_at: int = Field(
+        description="Epoch milliseconds of the last activity the orchestrator recorded for this server",
+        ge=0,
+    )
+    idle_seconds: float = Field(description="Seconds since 'last_activity_at'", ge=0)
+    pod_phase: Optional[str] = Field(
+        default=None, description="Kubernetes phase of the server pod, or null when no pod matches"
+    )
+    pod_ready: bool = Field(default=False, description="True when the pod is Running with all containers ready")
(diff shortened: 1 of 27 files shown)


vpoliakov-pixel:
## Manual verification — migration 004 and the reaper (needs Docker / a cluster)

I could not run these here: no Docker daemon on this machine, and the workflow I followed
forbids changing host infrastructure. Recipes below; they double as the regression recipe next
time `keepalive_until` or the reaper changes.

### 1. Migration 004 round-trip (Docker only)

```sh
uv run pytest integration-tests/database/test_migration_roundtrip.py -q
uv run pytest integration-tests/database/test_watcher_cleanup.py -q
```

**Expect:** all pass. The round-trip walks base → 004 → base one revision at a time with rows
seeded at each step, and asserts the head schema matches the ORM models. `seed_revision_004`
writes `keepalive_until = 1`, so the downgrade is exercised against a populated column.

**Failure looks like:** `column "keepalive_until" of relation "servers" does not exist` (the
model and the migration disagree), or a `test_head_schema_matches_the_orm_models` mismatch
listing `keepalive_until` (added to one side only).

### 2. Single head

```sh
uv run alembic -c orchestrator/src/idegym/orchestrator/alembic.ini heads
```

**Expect:** exactly `004 (head)`. Two lines means someone else landed an 004 concurrently and
this one needs renumbering to 005 — exact-revision rollback needs the history linear.

### 3. The reaper honours a hold (cluster)

```sh
uv run pytest e2e-tests/test_watcher.py -m e2e -q
```

**Expect:** both tests pass. They insert the *same* always-stale server row; the only
difference is `keepalive_until`, so one is reaped to `KILLED` and the other stays `ALIVE`.

**Failure looks like:** `Watcher reaped a server held by keepalive (status: 'KILLED')` — the
watcher is not reading the column, or the hold is being compared against the wrong clock.

### 4. Status and listing (cluster)

```sh
uv run pytest e2e-tests/test_server_lifecycle.py e2e-tests/test_server_strategies.py -m e2e -q
```

**Expect:** status answers `usable=False, availability=STOPPED` *after* the server is stopped
rather than raising; a second registration sees an empty `list_servers()`; and `reused` is
`False → True → False` across the start/finish/start/stop/start sequence.

**Failure looks like:** a `410 Gone` out of `server.status()` — the handler regressed to
`validate_server` instead of the ownership-only lookup.

samorojy (commented):
Review notes below: 4 correctness issues, 4 cleanups, 1 logging-convention issue and 1 test gap. The correctness ones (keepalive race, stale keepalive on reuse, pods[0] in status, status failing on Kubernetes errors) matter most.

orchestrator/src/idegym/orchestrator/database/database.py
```diff
@@ -471,6 +471,26 @@ async def save_idegym_server(
...
+    by keepalive would contradict whatever put it in that state.
+    """
+    server = await get_idegym_server(db, server_id)
+    if not server:
+        return None
+
+    if AvailabilityStatus(server.availability).is_terminal:
+        return server
+
+    server.keepalive_until = max(server.keepalive_until or 0, until)
```
samorojy: **Lost update: concurrent keepalives can shorten the hold.** This is a read-modify-write on an unlocked row, so the "never shortened" guarantee in the docstring doesn't hold under concurrency.

Example: holder A sends `minutes=60` and holder B sends `minutes=15` at the same time. Both read `keepalive_until=NULL`, A commits now+60m, then B commits now+15m. The row ends at now+15m and the watcher reaps the server 45 minutes before A's hold expires.

Suggestion: do it atomically, e.g. `UPDATE servers SET keepalive_until = GREATEST(COALESCE(keepalive_until, 0), :until) WHERE id = :id AND availability NOT IN (...terminal...) RETURNING ...`, or `SELECT ... FOR UPDATE` first.

orchestrator/src/idegym/orchestrator/database/models.py
```diff
@@ -59,6 +59,11 @@ class IdeGYMServer(Base):
     # for servers started from a snapshot, otherwise the server's own generated_name.
     snapshot_id = Column(String, index=True, nullable=True)
 
+    # Epoch millis until which an explicit keepalive holds this server against the inactivity reaper.
+    # Distinct from last_heartbeat_time, which only ever means "a request finished at this point":
+    # a sandbox can be legitimately in use and quiet, and only the client knows that.
+    keepalive_until = Column(BigInteger, nullable=True)
```
samorojy: **`keepalive_until` is never cleared on finish or reuse.** `update_idegym_server_owner` (database.py) only swaps `client_id`, so a reused server keeps the previous owner's hold.

Example: client A calls `keepalive(minutes=1440)`, then `finish_server`. Client B's RESTART/RESET start reuses the FINISHED server. B now owns a server the watcher won't reap for up to 24h however idle B is, and `list_servers`/status show B a `keepalive_until` it never asked for. Before the reuse, the same stale hold also keeps a FINISHED server alive past `finished_timeout`.

Suggestion: reset `keepalive_until = None` on finish and in `update_idegym_server_owner`.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -721,6 +721,22 @@ async def list_pods(label_selector: str, namespace: str) -> list[V1Pod]:
...
+    """Return the phase of the server's pod and whether every container in it is ready.
+
+    Terminating pods are skipped, so a restart in progress reports the incoming pod rather than
+    the one on its way out. The phase is ``None`` when no pod matches the selector at all.
+    """
+    pods = [pod for pod in await list_pods(label_selector, namespace) if pod.metadata.deletion_timestamp is None]
+    if not pods:
+        return None, False
+
+    pod = pods[0]
```
samorojy: **`pods[0]` can report the wrong pod.** The server is backed by a Deployment, which can have several non-terminating pods at once.

Example: the server pod is evicted. The evicted pod stays in phase `Failed` with no `deletion_timestamp`, and the Deployment starts a replacement that is Running and ready. If the evicted pod is listed first, `GET /status` returns `pod_phase='Failed'`, `pod_ready=False` for a healthy server.

Suggestion: prefer a Running/ready pod, or aggregate across pods the way `pods_are_ready` does.

orchestrator/src/idegym/orchestrator/router/server.py
```diff
@@ -177,6 +187,104 @@ async def get_server_capabilities(server_id: int, client_id: UUID, low_level_req
...
+async def get_server_status(server_id: int, client_id: UUID) -> ServerStatusResponse:
+    """Report whether a server is usable, without touching it.
+
+    Before this existed the cheapest liveness probe was an unrelated endpoint such as
+    ``/capabilities``, which happens to touch both the database record and the pod. This reads
+    the record and the pod phase directly, and deliberately does not update the server's
+    activity timestamp, so polling it cannot keep a server from being reaped.
+    """
+    server = await get_owned_server(client_id=client_id, server_id=server_id)
+    pod_phase, pod_ready = await pod_phase_and_readiness(f"app={server.generated_name}", server.namespace)
```
samorojy: **Status fails completely if the Kubernetes call fails.** The Kubernetes API is called on every status request, and an error from `list_pods` (API timeout, RBAC denial, deleted namespace) goes through `handle_server_exceptions` and becomes a 500. The client then loses `availability`/`details` too, although the DB already has them. That's the opposite of what an endpoint for reporting stopped/crashed servers should do. For terminal servers the pod lookup is also wasted work on every poll.

Suggestion: skip the pod call when availability is terminal, and return `pod_phase=None` / `pod_ready=None` on a Kubernetes error instead of failing the request.

orchestrator/src/idegym/orchestrator/database/helpers.py
```diff
@@ -87,6 +88,48 @@ async def update_client_status(db: AsyncSession, client_id: UUID, availability_s
     return client
 
 
+@with_db_session
+async def extend_server_keepalive(db: AsyncSession, client_id: UUID, server_id: int, until: int):
+    """Hold a client's own server against the inactivity reaper until ``until`` epoch millis."""
+    await _load_owned_server(db=db, client_id=client_id, server_id=server_id)
```
samorojy: **Redundant round-trips and an unreachable branch.** `_load_owned_server` already loads the client and server, then `extend_idegym_server_keepalive` selects the server again and refreshes it. That's 4 DB round-trips per keepalive. The 404 branch after it can never run, because `_load_owned_server` has already raised.

A single conditional `UPDATE ... RETURNING` with an ownership predicate would do the job and also fix the lost-update race (see the comment on database.py).

orchestrator/src/idegym/orchestrator/database/helpers.py
```diff
@@ -87,6 +88,48 @@ async def update_client_status(db: AsyncSession, client_id: UUID, availability_s
...
+async def get_owned_server(db: AsyncSession, client_id: UUID, server_id: int):
+    """Look up a server and check the client owns it, whatever state the server is in.
+
+    Unlike ``validate_server`` this accepts a finished, stopped or crashed server: an endpoint
+    that reports status has to be able to report exactly those.
+    """
+    return await _load_owned_server(db=db, client_id=client_id, server_id=server_id)
+
+
+async def _load_owned_server(db: AsyncSession, client_id: UUID, server_id: int):
```
samorojy: **Duplicated ownership checks.** `_load_owned_server` copies the client/server/ownership checks from `validate_server` line for line. Suggest having `validate_server` call `_load_owned_server` and then add only its usability check. Otherwise a later change to the ownership rules or error messages in one copy leaves status/keepalive enforcing different rules from forwarding/stop/finish.

api/src/idegym/api/orchestrator/servers.py
```diff
@@ -218,6 +255,66 @@ class ServerRequestResponse(BaseModel):
...
+    )
+    details: Optional[str] = Field(default=None, description="Failure reason recorded on a terminal status")
+
+
+class ListServersResponse(BaseModel):
+    client_id: UUID
+    servers: list[ServerSummary] = Field(default_factory=list)
+
+
+class ServerStatusResponse(BaseModel):
```
samorojy: **Duplicated model fields and usability rule.** `ServerStatusResponse` redefines all 11 fields of `ServerSummary` instead of extending it. The "usable" check (`availability in {ALIVE, REUSED}`) is also written out separately in `list_servers`, `get_server_status` and `validate_server`.

Suggestion: `class ServerStatusResponse(ServerSummary)` plus an `AvailabilityStatus.is_usable` property, so adding a field or changing what counts as usable is a single edit and list/status can't disagree.

orchestrator/src/idegym/orchestrator/database/helpers.py
```diff
@@ -191,6 +234,15 @@ async def find_matching_finished_server_in_db(
     return lookup_result.server, client_name
 
 
+@with_db_session
+async def list_client_servers(db: AsyncSession, client_id: UUID, include_terminal: bool):
+    """Return every server owned by a client, newest first, optionally including dead ones."""
+    servers = await get_idegym_servers_by_client_id(db, client_id)
+    if not include_terminal:
+        servers = [server for server in servers if not AvailabilityStatus(server.availability).is_terminal]
+    return sorted(servers, key=lambda server: server.created_at or 0, reverse=True)
```
samorojy: **Filtering and sorting in Python.** `list_client_servers` loads every server the client has ever owned, then drops terminal ones and sorts in Python. A long-running client with thousands of historical servers pays a full scan and hydration on every `list_servers(include_terminal=False)` just to get a few live rows.

Suggestion: filter on `availability` and `ORDER BY created_at DESC` in the query. (The router also opens a separate session just for `validate_client`, which could share the same one.)

watcher/src/idegym/watcher/cleanup.py
```diff
@@ -70,6 +70,15 @@ async def cleanup_servers(db: AsyncSession, current_time: int, inactive_timeout:
             f"(active: {time_since_last_heartbeat / 1000 / 60:.2f} minutes ago)"
         )
 
+        # An explicit keepalive is the client saying it still holds this server. Request activity
+        # is only a proxy for that, and a bad one while an agent thinks or a build runs.
+        if server.keepalive_until and server.keepalive_until > current_time:
+            logger.debug(
+                f"Skipping IdeGYM server {server.generated_name}: held by keepalive for another "
```
samorojy: **Logging convention.** Per AGENTS.md: "Bind context as keyword arguments (`logger.info("Server started", server=name)`) rather than formatting into the message." Here the server name and remaining minutes are formatted into an f-string, so structured log queries can't filter held servers by `server` or `keepalive_until`.

integration-tests/database/test_migration_roundtrip.py
```diff
@@ -177,7 +197,7 @@ async def test_round_trip_through_every_revision(manager: MigrationManager):
 
 
 async def test_downgrading_only_the_last_revision_preserves_rows(manager: MigrationManager):
```
samorojy: **Test no longer covers a one-step rollback.** This still runs head→002→003, so now that 004 is head it downgrades two revisions. Nothing checks that the 004→003 rollback (the one `scripts/rollback.py` actually runs) keeps rows and only drops `keepalive_until`, so a broken `004_down.sql` wouldn't be caught. Suggest updating it to downgrade head→003 and assert the rows survive.

samorojy (approved):

PR #294, opened 2026-08-31
[JBRes-10676] SDK: raising failures, injectable HTTP client, loop-owning facade (stack 6/8)

vpoliakov-pixel:
**Stack 6/8** — #289 ← #290 ← #291 ← #292 ← #293 ← **#294** ← #295 ← #296 (merge in order, oldest first).
Base: `…-05-…`. Branch: `vladimir.poliakov/jbres-10676-06-…`.

Improvements #12, #13 and #14 of JBRes-10676 — the Medium SDK items. Sits above stack #3 because `stop_server` raises the typed exceptions introduced there.

## What changed

**Stopping a server raises on failure** (#12). A failed delete came back as an `ErrorResponse` *return value*, unlike the rest of the API, so a caller who did not check recorded a live pod as stopped and leaked it. Both methods already declared `-> ServerActionResponse` while able to return `ErrorResponse`, so the type was a lie either way. `snapshot_server` is untouched — its signature declares the union honestly.

**The HTTP client can be supplied or configured** (#13). An integration needing a different HTTP stack or bounded pooling had to reach into `client._http_client._transport` after construction. `transport` and `limits` configure the client IdeGYM builds; `http_client` is the escape hatch, used verbatim and **not closed on exit** because it belongs to its caller. Combining them raises rather than silently ignoring.

**A loop-owning facade** (#14). `IdeGYMClient` is bound to the loop that created it, so a caller driving sandboxes from several loops cannot share one registration. That matters beyond convenience: deregistering terminates every server a client owns, so two registrations in one process means either can tear down the other's sandboxes. `SharedIdeGYMClient` owns a loop in a dedicated thread and marshals calls onto it — roughly the 150 lines each such integration writes for itself.

## How to verify

```
uv run pytest -m unit unit-tests/test_client_exceptions.py \
  unit-tests/test_client_http_configuration.py unit-tests/test_shared_client.py
```

- The stop/restart tests guard a regression that is silent by construction — the old code *returned* the failure, so a happy-path-only test would pass while a pod leaked.
- The facade tests use a real thread pool and a second event loop: 32 calls from 8 threads share one registration, and a caller already running its own loop reaches the same one.
- Ownership is pinned both ways: a built client is closed on exit, a supplied one is not.

Full suite: **1297 passed, 2 skipped**; docs site builds.

Resolves: JBRes-10676 (partially — #12, #13, #14)

### DIFF
--- a/client/src/idegym/client/client.py
+++ b/client/src/idegym/client/client.py
@@ -138,7 +138,11 @@ class IdeGYMClient:
                 base_url=orchestrator_url,
                 timeout=request_timeout_in_seconds,
                 transport=transport,
-                limits=limits if limits is not None else Limits(),
+                # Only override when asked: a bare `Limits()` is not httpx's default. It sets
+                # max_connections=None, which removes the 100-connection pool cap entirely, so
+                # passing it unconditionally would unbound the pool for every caller that never
+                # asked to configure one.
+                **({"limits": limits} if limits is not None else {}),
                 headers=(
                     {
                         "Authorization": f"Basic {credential}",
--- a/client/src/idegym/client/client.py
+++ b/client/src/idegym/client/client.py
@@ -62,6 +62,11 @@ class ServerCloseAction(StrEnum):
 class IdeGYMClient:
     """
     HTTP client for interacting with the IdeGYM orchestrator and server APIs.
+
+    **This object is bound to the event loop that created it.** It owns an ``httpx`` session and
+    a heartbeat task, so every call has to be awaited on that same loop. If you drive sandboxes
+    from more than one loop, use :class:`~idegym.client.shared.SharedIdeGYMClient`, which owns a
+    loop in its own thread and lets any caller share one registration.
     """
 
     def __init__(
--- a/client/src/idegym/client/client.py
+++ b/client/src/idegym/client/client.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from typing import Optional
 from uuid import UUID, uuid4
 
-from httpx import AsyncClient
+from httpx import AsyncBaseTransport, AsyncClient, Limits
 from idegym.api.auth import BasicAuth
 from idegym.api.config import OTELConfig, TracingConfig
 from idegym.api.health import HealthCheckResponse
@@ -75,6 +75,9 @@ class IdeGYMClient:
         heartbeat_interval_in_seconds: int = 60,
         request_timeout_in_seconds: int = 60,
         otel_config: Optional[OTELConfig] = None,
+        transport: Optional[AsyncBaseTransport] = None,
+        limits: Optional[Limits] = None,
+        http_client: Optional[AsyncClient] = None,
     ):
         """
         Initialize the IdeGYM HTTP client.
@@ -94,6 +97,17 @@ class IdeGYMClient:
             otel_config: OpenTelemetry configuration for tracing. Falls back to ``IDEGYM_OTEL_*``
                 environment variables when not provided. Tracing stays off unless an endpoint is
                 configured, either here or through ``IDEGYM_OTEL_TRACING_ENDPOINT``.
+            transport: Transport for the HTTP client this object builds — for an alternative HTTP
+                stack, a recording transport in tests, or a proxy.
+            limits: Connection-pool limits for the HTTP client this object builds.
+            http_client: A fully configured ``httpx.AsyncClient`` to use verbatim. Nothing about it
+                is modified, so it must already carry ``base_url`` and any authentication, and it
+                is not closed on exit — its owner closes it. Mutually exclusive with ``transport``
+                and ``limits``.
+
+        Raises:
+            ValueError: if ``http_client`` is combined with ``transport`` or ``limits``, since the
+                supplied client is used as-is and those arguments would be silently ignored.
         """
         if orchestrator_url == "idegym.test":
             orchestrator_url = f"http://{orchestrator_url}"
@@ -107,20 +121,30 @@ class IdeGYMClient:
         if not orchestrator_url == "http://idegym.test" and not (auth.username and auth.password):
             raise ValueError("Username and password must be provided or set in environment variables")
 
-        http_client = AsyncClient(
-            base_url=orchestrator_url,
-            timeout=request_timeout_in_seconds,
-            headers=(
-                {
-                    "Authorization": f"Basic {credential}",
-                    "Content-Type": "application/json",
-                }
-                if (credential := auth.base64)
-                else {
-                    "Content-Type": "application/json",
-                }
-            ),
-        )
+        if http_client is not None and (transport is not None or limits is not None):
+            raise ValueError(
+                "transport and limits configure the client IdeGYM builds; they do not apply to http_client"
+            )
+
+        # A supplied client belongs to its caller: used as-is, and closed by them, not here.
+        owns_http_client = http_client is None
+        if http_client is None:
+            http_client = AsyncClient(
+                base_url=orchestrator_url,
+                timeout=request_timeout_in_seconds,
+                transport=transport,
+                limits=limits if limits is not None else Limits(),
+                headers=(
+                    {
+                        "Authorization": f"Basic {credential}",
+                        "Content-Type": "application/json",
+                    }
+                    if (credential := auth.base64)
+                    else {
+                        "Content-Type": "application/json",
+                    }
+                ),
+            )
 
         # Tracing is opt-in: with no endpoint the exporter is never built, so a caller who
         # does not know about OTEL cannot end up shipping telemetry off their infrastructure.
@@ -142,6 +166,7 @@ class IdeGYMClient:
         )
 
         self._http_client: AsyncClient = http_client
+        self._owns_http_client: bool = owns_http_client
         self._otel_config: OTELConfig = otel_config
 
         self._heartbeat_interval_in_seconds: int = heartbeat_interval_in_seconds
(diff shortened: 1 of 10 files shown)


client/src/idegym/client/shared.py
```diff
@@ -0,0 +1,155 @@
...
+            logger.debug("Loop drain did not finish cleanly; stopping anyway", exc_info=True)
+        self._loop.call_soon_threadsafe(self._loop.stop)
+        self._thread.join()
+        self._loop.close()
+
+    async def _drain(self) -> None:
+        """Let cancellations be delivered, then run the loop's own shutdown hooks."""
+        pending = [task for task in asyncio.all_tasks(self._loop) if task is not asyncio.current_task()]
+        if pending:
+            await asyncio.wait(pending, timeout=self._DRAIN_TIMEOUT_SECONDS)
```
samorojy: **Pending caller tasks are never cancelled, so a caller thread can hang forever on shutdown.**

`_drain` waits up to 5s for pending tasks but doesn't cancel them. `close()` then stops and closes the loop. If another thread is blocked in `run()`/`submit()` with the default `timeout=None` (e.g. `shared.run(lambda c: server.execute_bash("sleep 60"))`), its `Future.result()` never returns and no error is raised. A `submit` that lands between `loop.stop` and `loop.close` hangs the same way.

Suggestion: cancel the remaining tasks after the wait (and await them), and reject `submit` once closing has started.

client/src/idegym/client/shared.py
```diff
@@ -0,0 +1,155 @@
...
+    def close(self) -> None:
+        """Drain the loop before stopping it, so nothing is torn down mid-flight.
+
+        ``IdeGYMClient.__aexit__`` cancels the heartbeat task without awaiting it, and the file
+        transfers hand work to ``asyncio.to_thread``. Stopping the loop immediately would leave
+        those pending — Python then prints "Task was destroyed but it is pending!" on every exit,
+        and the executor and async generators never get their shutdown hooks.
+        """
+        try:
+            self.submit(self._drain).result(timeout=self._DRAIN_TIMEOUT_SECONDS)
```
samorojy: **The outer timeout is the same as the inner one, so the outer always expires first.**

`.result(timeout=_DRAIN_TIMEOUT_SECONDS)` here matches `asyncio.wait(..., timeout=_DRAIN_TIMEOUT_SECONDS)` inside `_drain`. When any task is still pending (e.g. a slow `asyncio.to_thread` file transfer), this `.result()` raises `TimeoutError` before `_drain` gets to `shutdown_asyncgens()` / `shutdown_default_executor()`. The loop is then stopped with `_drain` itself pending, which produces the "Task was destroyed but it is pending!" warning this code is meant to prevent.

Suggestion: give the outer wait extra headroom over the inner one (or bound the shutdown hooks separately).

client/src/idegym/client/shared.py
```diff
@@ -0,0 +1,155 @@
...
+        client = self.client
+        return self._loop.submit(lambda: call(client))
+
+    def run[T](self, call: Callable[[IdeGYMClient], Awaitable[T]], timeout: Optional[float] = None) -> T:
+        """Run ``call(client)`` on the owned loop and return its result.
+
+        Safe to call from any thread, including one that is itself running an event loop — the
+        awaitable never touches the caller's loop.
+        """
+        return self.submit(call).result(timeout)
```
samorojy: **A timed-out `run()` doesn't cancel the underlying work.**

When `Future.result(timeout)` raises `TimeoutError`, the coroutine keeps running on the owned loop. For example, `shared.run(lambda c: c.start_server(...), timeout=10)` can time out and still create a server the caller never got a handle to. It then shows up as a pending task at exit (see the `close()`/`_drain` comments).

Suggestion: on `TimeoutError`, call `future.cancel()` before re-raising.

client/src/idegym/client/shared.py
```diff
@@ -0,0 +1,155 @@
...
+        if self._client is None:
+            raise RuntimeError("SharedIdeGYMClient is not started; use it as a context manager")
+        return self._client
+
+    def __enter__(self) -> Self:
+        self._loop = _LoopThread(name="idegym-client-loop")
+
+        async def start() -> IdeGYMClient:
+            client = IdeGYMClient(**self._client_arguments)
+            await client.__aenter__()
```
samorojy: **A failed `__aenter__()` leaks the `IdeGYMClient`'s httpx client.**

If `client.__aenter__()` raises here (e.g. registration is refused), nothing closes or uninstruments the `AsyncClient` the constructor just built. `__enter__` only closes the loop. That abandons a connection pool on a closed loop: an unclosed-client `ResourceWarning` and leaked sockets on every failed entry.

Suggestion: on failure, `await client.__aexit__(*sys.exc_info())` (or at least `aclose()` the http client) before re-raising.

samorojy (commented):
Review notes, most important first: the `SharedIdeGYMClient.close()` / `_drain` shutdown path (can hang a caller thread, and the equal timeouts defeat the drain) and `stop_server` raising inside `with_server`'s `finally`. The `limits`/`transport` and `http_client` credential cases contradict the PR's own docs.

One more that's outside the diff: `IdeGYMClient.start_server` (`client.py` ~L435) still maps `ErrorResponse` to `http_error` inline, duplicating the new `raise_for_error_response`. It could call `raise_for_error_response(server_response, "Starting server")` so the two message formats don't drift.

client/src/idegym/client/client.py
```diff
@@ -107,20 +126,34 @@ def __init__(
...
+            raise ValueError(
+                "transport and limits configure the client IdeGYM builds; they do not apply to http_client"
+            )
+
+        # A supplied client belongs to its caller: used as-is, and closed by them, not here.
+        owns_http_client = http_client is None
+        if http_client is None:
+            http_client = AsyncClient(
+                base_url=orchestrator_url,
+                timeout=request_timeout_in_seconds,
```
samorojy: **`limits` is silently ignored when a custom `transport` is supplied.**

httpx's `_init_transport` returns a supplied `transport` as-is, so `limits` only applies when httpx builds the transport itself. This check only rejects `limits` combined with `http_client`, and the docs example (`website/docs/reference/client.md` ~L97-98) passes `limits=` and `transport=` together. That example quietly loses the pool cap, which is the silently-ignored case this `ValueError` was added to prevent.

Suggestion: also reject `transport` + `limits`, and fix the docs example.

client/src/idegym/client/client.py
```diff
@@ -107,20 +126,34 @@ def __init__(
         if not orchestrator_url == "http://idegym.test" and not (auth.username and auth.password):
```
samorojy: **Credentials are still required when `http_client` is supplied.**

The docs say a supplied `http_client` "must already carry base_url and any authentication", but this check still runs. A caller who passes `http_client=AsyncClient(base_url=..., auth=...)` without `IDEGYM_AUTH_USERNAME/PASSWORD` in the environment gets `ValueError("Username and password must be provided...")`. `orchestrator_url`, `auth` and `request_timeout_in_seconds` are also silently ignored in that case.

Suggestion: skip the credential check when `http_client` is given, and reject (or document as ignored) the arguments that no longer apply.

client/src/idegym/client/client.py
```diff
@@ -216,7 +250,8 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
             client=self._http_client,
             config=self._otel_config,
         )
-        await self._http_client.aclose()
+        if self._owns_http_client:
```
samorojy: **OTEL still instruments and uninstruments a caller-owned `http_client`.**

`aclose()` is now skipped for a supplied client, but `instrument()` in `__init__` and `uninstrument()` just above still modify its transport. That contradicts "Nothing about it is modified" in the docs, and it breaks sharing: with two `IdeGYMClient`s on one pooled client, the second `instrument` is skipped as already instrumented, and when the first exits it strips tracing from the client that the second (and the owner) are still using.

Suggestion: only instrument and uninstrument when `self._owns_http_client`.

client/src/idegym/client/operations/servers.py
```diff
@@ -166,19 +167,26 @@ async def stop_server(
...
             response_raw=response_raw, model_class=ServerActionResponse
         )
-        return await self._utils.wait_for_async_operation_to_end(
+        result = await self._utils.wait_for_async_operation_to_end(
             operation_id=response.operation_id,
             success_response_model=ServerActionResponse,
             error_response_model=ErrorResponse,
             polling_config=polling_config,
         )
+        return raise_for_error_response(result, f"Stopping server {server_id}")
```
samorojy: **Raising here changes how `stop_server` behaves through retry and `with_server`.**

1. `IdeGYMClient.stop_server` is wrapped in `@retry_with_backoff(attempts=3)`, so permanent 4xx failures (e.g. 410 because the pod is already gone) now re-send the DELETE three times with backoff.
2. `with_server(close_action=STOP)` calls `stop_server` in its `finally`. If the body raised, the cleanup exception replaces the original one, which survives only as `__context__`.

Suggestion: don't retry non-retryable statuses, and in `with_server` log or suppress cleanup errors while an exception is already propagating.

client/src/idegym/client/exceptions.py
```diff
@@ -114,3 +115,19 @@ def http_error(
 ) -> IdeGYMHTTPError:
     """Build the most specific exception for ``status_code``, ready to raise."""
     return error_class_for_status(status_code)(message, status_code=status_code, body=body, method=method, url=url)
+
+
+def raise_for_error_response[T](response: T | ErrorResponse, operation: str) -> T:
```
samorojy: **Converting `ErrorResponse` per operation leaves other operations silent.**

`snapshot_server`, `register_client` and `stop_client` still *return* `ErrorResponse`. `IdeGYMClient.__aexit__` discards the `_stop_client()` result, so a failed deregistration leaves every server the client owns running with no error raised. That is the same pod-leak class this PR targets.

Suggestion: raise inside `HTTPUtils.wait_for_async_operation_to_end` when `error_response_model` matches, which covers every operation at once.

samorojy (approved):

PR #295, opened 2026-08-31
[JBRes-10676] Per-server labels and annotations (stack 7/8)

vpoliakov-pixel:
**Stack 7/8** — #289 ← #290 ← #291 ← #292 ← #293 ← #294 ← **#295** ← #296 (merge in order, oldest first).
Base: `…-06-…`. Branch: `vladimir.poliakov/jbres-10676-07-…`.

Improvement #17 of JBRes-10676, which the ticket calls the highest ratio of operator value to effort on the list. Its own PR because it is self-contained and touches all five plumbing layers; it sits here only to avoid conflicting with the client signatures the SDK PRs below change.

## What changed

`pod_overrides` is a partial `V1PodSpec`, which cannot set object metadata, and there was no labels field — so a sandbox could not be tagged with the job, team or task that created it, and none of `kubectl get pods -l …`, cost attribution or dashboard grouping applied. The workaround was folding metadata into the server *name*, against a 63-character RFC-1035 budget.

`StartServerRequest.labels` and `.annotations` are now applied to the Deployment, Pod, Service and PodDisruptionBudget.

Two design points worth review attention:

- **Caller metadata merges underneath the managed keys**, so a managed key always wins, *and* the request model rejects any attempt to set one (`app`, `app.kubernetes.io/*`, `idegym.jetbrains.com/*`) rather than silently overwriting. Those are what the Service selector, the PDB and the watcher's pod queries match on — taking one over would detach the sandbox from the machinery that manages it. The deploy layer enforces this independently of the validator, so the platform stays correct even if a request reaches it another way.
- **Neither field enters the snapshot hash.** They describe who asked for a sandbox, not what is in it, and a per-job label would otherwise make every snapshot a miss. They are likewise not part of reuse matching, so a reused server keeps the labels it was started with — documented, not fixed.

## How to verify

```
uv run pytest -m unit unit-tests/test_server_labels.py
```

- Deploy-layer tests assert the collision behaviour and that **no caller label ever leaks into a selector** — a selector that grew one would stop matching pods started without it.
- Request-model tests separate keys IdeGYM owns from keys that merely resemble them (`appliance`, `idegym.example.com/task`).
- `e2e-tests/test_pod_spec_overrides.py` does the thing labels exist for: finds the sandbox by selector without knowing its generated name. Collected only, no cluster here.

Full suite: **1318 passed, 2 skipped**; docs site builds.

Resolves: JBRes-10676 (partially — #17)

### DIFF
--- a/api/src/idegym/api/orchestrator/servers.py
+++ b/api/src/idegym/api/orchestrator/servers.py
@@ -9,8 +9,19 @@ from idegym.api.pod_spec import (
     KubernetesVolumeMount,
 )
 from idegym.api.resources import KubernetesResources
-from idegym.api.type import KubernetesNodeSelector, KubernetesObjectName, OCIImageName
-from pydantic import BaseModel, Field
+from idegym.api.type import (
+    KubernetesAnnotations,
+    KubernetesLabels,
+    KubernetesNodeSelector,
+    KubernetesObjectName,
+    OCIImageName,
+)
+from pydantic import BaseModel, Field, field_validator
+
+# Labels IdeGYM sets itself: the Service selector, the PodDisruptionBudget and the watcher's
+# pod queries all match on these, so a caller must not be able to take them over.
+MANAGED_LABEL_KEYS = frozenset({"app"})
+MANAGED_LABEL_PREFIXES = ("app.kubernetes.io/", "idegym.jetbrains.com/")
 
 
 class ServerReuseStrategy(StrEnum):
@@ -94,6 +105,25 @@ class StartServerRequest(BaseModel):
         description="Kubernetes node selector labels for scheduling the server pod",
         examples=[{"kubernetes.io/os": "linux"}],
     )
+    labels: KubernetesLabels = Field(
+        default_factory=dict,
+        description=(
+            "Extra labels applied to the server's Deployment, Pod, Service and PodDisruptionBudget, "
+            "so a sandbox can be found with 'kubectl get pods -l ...' and grouped for cost "
+            "attribution. Keys IdeGYM manages ('app', 'app.kubernetes.io/*', 'idegym.jetbrains.com/*') "
+            "are rejected rather than silently ignored, since overwriting them would break the "
+            "selectors that address the pod."
+        ),
+        examples=[{"team": "research", "job": "swe-bench-run-42"}],
+    )
+    annotations: KubernetesAnnotations = Field(
+        default_factory=dict,
+        description=(
+            "Extra annotations applied to the server pod. Use for metadata too long or too "
+            "unstructured to be a label, such as a task URL or a serialized request id."
+        ),
+        examples=[{"idegym.example.com/task-url": "https://tracker.example.com/TASK-1"}],
+    )
     volumes: list[KubernetesVolume] = Field(
         default_factory=list,
         description="Pod-level volumes (native Kubernetes shape), mounted into the server container via 'volume_mounts'",
@@ -169,6 +199,20 @@ class StartServerRequest(BaseModel):
         ),
         examples=[0, 3],
     )
+
+    @field_validator("labels")
+    @classmethod
+    def _reject_managed_label_keys(cls, labels: KubernetesLabels) -> KubernetesLabels:
+        """Refuse to accept a label IdeGYM owns rather than accepting and then overwriting it.
+
+        The managed labels are what the Service selector, the PodDisruptionBudget and the
+        watcher's pod queries match on, so a caller who overwrote one would detach their own
+        sandbox from the machinery that manages it.
+        """
+        reserved = sorted(key for key in labels if key in MANAGED_LABEL_KEYS or key.startswith(MANAGED_LABEL_PREFIXES))
+        if reserved:
+            raise ValueError(f"labels may not set IdeGYM-managed keys: {', '.join(reserved)}")
+        return labels
 
 
 class ServerScopedRequest(BaseModel):
--- a/api/src/idegym/api/type.py
+++ b/api/src/idegym/api/type.py
@@ -59,7 +59,12 @@ KubernetesLabelValue = Annotated[
 ]
 
 
+# An annotation key follows the label-key syntax; the value is arbitrary and may be long.
+KubernetesAnnotationKey: TypeAlias = KubernetesLabelKey
+
 KubernetesNodeSelector: TypeAlias = dict[KubernetesLabelKey, KubernetesLabelValue]
+KubernetesLabels: TypeAlias = dict[KubernetesLabelKey, KubernetesLabelValue]
+KubernetesAnnotations: TypeAlias = dict[KubernetesAnnotationKey, str]
 AuthType: TypeAlias = Literal["Basic", "Bearer", "Token"]
 Duration: TypeAlias = timedelta
 LogLevel: TypeAlias = int
--- a/backend-utils/src/idegym/backend/utils/kubernetes_client.py
+++ b/backend-utils/src/idegym/backend/utils/kubernetes_client.py
@@ -269,6 +269,8 @@ async def deploy_server(
     server_kind: ServerKind = ServerKind.IDEGYM,
     snapshot_id: Optional[str] = None,
     snapshot_tag: Optional[str] = None,
+    extra_labels: Optional[dict[str, str]] = None,
+    extra_annotations: Optional[dict[str, str]] = None,
 ):
     """
     Create a Kubernetes Deployment, Service, and PodDisruptionBudget for a server.
@@ -327,7 +329,10 @@ async def deploy_server(
     )
 
     image_pull_secret = V1LocalObjectReference(name="regcred")
+    # Caller metadata goes on first so a managed key always wins: the selectors that address the
+    # pod are built from the managed labels, and prometheus scraping from the managed annotations.
     annotations = {
+        **(extra_annotations or {}),
         "cluster-autoscaler.kubernetes.io/safe-to-evict": "false",
         **prometheus_annotations,
     }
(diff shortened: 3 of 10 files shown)


vpoliakov-pixel:
## Manual verification — labels reach the pod (needs a cluster)

Unit tests assert what `deploy_server` *builds*; only a cluster shows that Kubernetes accepts
it and that the selector workflow works. Not run here — no minikube on this machine.

```sh
uv run pytest e2e-tests/test_pod_spec_overrides.py -m e2e -q
```

**Expect:** 5 passed, including the two new ones. `test_labels_and_annotations_reach_the_pod`
starts a server with `labels={"idegym-e2e-job": …, "team": "research"}` and then finds it
**by selector, without knowing its generated name** — which is the whole point of the feature.

Or by hand, against a running server started with those labels:

```sh
kubectl get pods -n idegym -l team=research
kubectl get pod -n idegym -l team=research \
  -o jsonpath='{.items[0].metadata.labels}{"\n"}{.items[0].metadata.annotations}'
```

**Expect:** exactly one pod; its labels contain both `team=research` and the untouched
`app.kubernetes.io/part-of=idegym`; its annotations contain the caller's key alongside
`cluster-autoscaler.kubernetes.io/safe-to-evict: "false"`.

**Failure looks like:**
- `app.kubernetes.io/part-of` missing or holding a caller value → the merge order inverted and
  a caller can now detach the pod from the watcher's queries.
- `safe-to-evict: "true"` → the same inversion on annotations; the sandbox becomes evictable.
- The pod starting at all with `labels={"app": "…"}` → the request-model validator is not
  running; that request must be rejected with `422`.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -327,7 +329,10 @@ async def deploy_server(
     )
 
     image_pull_secret = V1LocalObjectReference(name="regcred")
+    # Caller metadata goes on first so a managed key always wins: the selectors that address the
+    # pod are built from the managed labels, and prometheus scraping from the managed annotations.
     annotations = {
+        **(extra_annotations or {}),
```
samorojy: **podsnapshot.gke.io/ps-name is only protected when a snapshot is requested.** The comment says a managed key always wins, but `ps-name` is only written when `snapshot_tag` is set. So a request with `snapshot=None` and `annotations={"podsnapshot.gke.io/ps-name": "<any snapshot in the namespace>"}` makes GKE restore the pod from that snapshot. That snapshot could belong to another tenant or image. Meanwhile the DB records `snapshot_id=None`, and the `snapshot-id` label is set to the server name. Please reject (or strip) reserved annotation prefixes like `podsnapshot.gke.io/`, in the same way labels are rejected.

api/src/idegym/api/type.py
```diff
@@ -59,7 +59,12 @@ class ConditionStatus(StrEnum):
 ]
 
 
+# An annotation key follows the label-key syntax; the value is arbitrary and may be long.
+KubernetesAnnotationKey: TypeAlias = KubernetesLabelKey
+
 KubernetesNodeSelector: TypeAlias = dict[KubernetesLabelKey, KubernetesLabelValue]
+KubernetesLabels: TypeAlias = dict[KubernetesLabelKey, KubernetesLabelValue]
```
samorojy: **`KubernetesLabelKey` accepts keys that Kubernetes rejects.** The regex puts no cap on the name segment at 63 characters (only the total at 317) and allows an uppercase DNS prefix. So `labels={"a"*100: "x"}` or `labels={"Example.COM/team": "x"}` passes validation and the request is accepted. Then `create_namespaced_deployment` fails with a 422, the server ends up `FAILED_TO_START`, and the caller sees a 500 instead of a validation error. `test_labels_are_held_to_the_kubernetes_syntax` does not cover these cases. Reusing this alias for caller-supplied labels makes it user-facing, so the name/prefix constraints should be tightened.

api/src/idegym/api/type.py
```diff
@@ -59,7 +59,12 @@ class ConditionStatus(StrEnum):
...
+KubernetesAnnotations: TypeAlias = dict[KubernetesAnnotationKey, str]
```
samorojy: **Annotation values have no size limit.** Kubernetes caps the total size of an object's annotations at 256 KiB. `KubernetesAnnotations = dict[..., str]` accepts any size, so an oversized payload is accepted, stored with the async operation request, and then fails at deploy time as a 500 / `FAILED_TO_START`. A validator on the total size would reject it up front.

orchestrator/src/idegym/orchestrator/router/server.py
```diff
@@ -484,6 +484,8 @@ async def _task_start_server(
                 server_kind=request.server_kind,
                 snapshot_id=request.snapshot.id if request.snapshot else None,
                 snapshot_tag=request.snapshot.tag if request.snapshot else None,
+                extra_labels=request.labels,
```
samorojy: **The snapshot pipeline does not forward labels or annotations.** The other `deploy_server` caller, `run_snapshot_pipeline_job` in `snapshot_pipeline.py`, takes the same `StartServerRequest` but never passes `extra_labels`/`extra_annotations`. So a prepare request with `labels={"team": "research"}` is accepted and the labels are silently dropped. The snapshot-prep pod, which uses real cluster resources, is left out of `kubectl get pods -l team=research` and out of cost attribution. Keeping them out of the hash is correct, but they should still be applied to the prep pod.

api/src/idegym/api/orchestrator/servers.py
```diff
@@ -94,6 +105,25 @@ class StartServerRequest(BaseModel):
         description="Kubernetes node selector labels for scheduling the server pod",
         examples=[{"kubernetes.io/os": "linux"}],
     )
+    labels: KubernetesLabels = Field(
```
samorojy: **With the default `reuse_strategy=RESET`, a new request's labels are dropped whenever a server is reused.** When `find_matching_finished_server_in_db` finds a matching server, the `existing_server` branch in `_task_start_server` never touches K8s metadata. The pod keeps the *previous* job's labels, so `kubectl get pods -l job=<new job>` returns nothing and cost goes to the old job. The docs mention this, but because RESET is the client default, the feature often won't take effect. Consider patching labels/annotations on the reused Deployment/pod, or at least documenting this in the field description.

api/src/idegym/api/orchestrator/servers.py
```diff
@@ -9,8 +9,19 @@
...
+    KubernetesLabels,
+    KubernetesNodeSelector,
+    KubernetesObjectName,
+    OCIImageName,
+)
+from pydantic import BaseModel, Field, field_validator
+
+# Labels IdeGYM sets itself: the Service selector, the PodDisruptionBudget and the watcher's
+# pod queries all match on these, so a caller must not be able to take them over.
+MANAGED_LABEL_KEYS = frozenset({"app"})
```
samorojy: **Two sources of truth for managed labels.** `MANAGED_LABEL_KEYS`/`MANAGED_LABEL_PREFIXES` repeats, by hand, the keys `deploy_server` builds in `match_labels`/`labels` (and `nodes_holder` builds as well). AGENTS.md even has to tell contributors to keep them in sync. If a new managed label is added in `deploy_server` without updating this constant, a caller can take it over and nothing fails. It would be more robust to define the managed labels once in `api/`, have `deploy_server` build from that, and have a unit test check that every label `deploy_server` emits is covered by the reservation.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -341,6 +346,7 @@ async def deploy_server(
         "app.kubernetes.io/part-of": "idegym",
     }
     labels = {
+        **(extra_labels or {}),
```
samorojy: **Managed annotations are overwritten silently, while managed labels are rejected.** A caller who sets `cluster-autoscaler.kubernetes.io/safe-to-evict: "true"` or `prometheus.io/*` gets no error, and their value is quietly replaced. The docs, meanwhile, promise that managed keys are "rejected, not silently overwritten". Please handle both the same way: reject managed annotation keys in the request model.

e2e-tests/test_pod_spec_overrides.py
```diff
@@ -105,3 +105,42 @@ async def test_pod_overrides_host_aliases(test_image, test_id):
...
+
+        pod = pods[0]
+        assert pod.metadata.labels["team"] == "research"
+        assert pod.metadata.annotations["idegym.example.com/task-url"] == "https://tracker.example.test/TASK-1"
+        # Managed labels are untouched, so the platform still addresses the pod.
+        assert pod.metadata.labels["app.kubernetes.io/part-of"] == "idegym"
+
+
+@pytest.mark.asyncio
+async def test_a_managed_label_is_rejected_before_anything_is_created(test_id, test_image):
```
samorojy: **This e2e test is really a pure unit test.** AGENTS.md says: *"Push coverage down. Anything expressible as a pure function belongs in `unit-tests/`. Reserve e2e for behaviour that genuinely requires a cluster."* `client.start_server` builds `StartServerRequest` on the client side, so the `ValidationError` is raised before any HTTP call and the orchestrator is never reached. `test_a_managed_label_key_is_rejected` in `unit-tests/test_server_labels.py` already covers this, yet the e2e version still registers a client against the cluster.

unit-tests/test_server_labels.py
```diff
@@ -0,0 +1,175 @@
...
+@pytest.fixture
+async def api_client():
+    client = ApiClient()
+    try:
+        yield client
+    finally:
+        await client.close()
+
+
+def _patch_clients(mocker, api_client):
```
samorojy: **Duplicated test helpers.** The `api_client` fixture and `_patch_clients` are copied almost line for line from `unit-tests/test_kubernetes_client.py` (lines 22-48). Two copies of the `create_clients` mock will drift apart. Consider moving them to a shared `unit-tests/conftest.py` fixture, or putting these `deploy_server` tests in `test_kubernetes_client.py`.

samorojy (approved):

PR #296, opened 2026-08-31
[JBRes-10676] Image source validation and honest start timeouts (stack 8/8)

vpoliakov-pixel:
**Stack 8/8** — #289 ← #290 ← #291 ← #292 ← #293 ← #294 ← #295 ← **#296** (merge in order, oldest first).
Base: `…-07-…`. Branch: `vladimir.poliakov/jbres-10676-08-…`.

Improvements #19 and #20 of JBRes-10676 — the ticket's own "Images and tooling" section. Both are about a build or start failing with a message that points at the wrong thing.

## What changed

**A source tree is checked before the renderer copies out of it** (#19). The `idegym-server` plugin copies a fixed set of paths out of a checkout, so a ref predating any of them — a commit copied from an example config older than the split into `plugins/` — failed deep inside the Docker build with a bare `cp: no such file`, naming neither the ref nor what was missing. That cost an afternoon during the live test.

The required paths are now one list feeding both the copies and the checks. A local root is validated on the host in `apply()`; a git ref is checked in the container immediately after the clone — as early as possible without the client talking to the git host. Either way the message names the source, the ref and **every** missing path at once.

The builder already publishes as `idegym-image-builder`, but nothing said so — which is why wrapping an image looked like it required a clone. Added an install section.

**The readiness default is raised and the failure named** (#20). Sixty seconds does not cover a cold pull of a multi-gigabyte image, and the failure read as a readiness timeout, pointing the reader at the image's health endpoint rather than at the pull. Both halves were wrong, so both change: the default is now **300s**, and the message says whether the pod was still pulling, unable to pull, or running with a readiness probe that has not passed. The kubelet has no distinct "Pulling" waiting reason — a pull in progress is `ContainerCreating` — which is exactly why the two were indistinguishable.

## How to verify

```
uv run pytest -m unit unit-tests/test_idegym_server_source_check.py \
  unit-tests/test_pod_startup_diagnosis.py
```

The source-check tests pin that the check precedes the first copy and that its loop covers exactly what the renderer copies — the drift that would quietly bring the bare `cp` failure back. The diagnosis tests assert wording, since a misleading message is the bug.

Full suite: **1337 passed, 2 skipped**; docs site builds; `uv lock --check` clean.

Resolves: JBRes-10676 (partially — #19, #20)


vpoliakov-pixel:
## Manual verification — the stale-ref failure message (needs Docker)

The unit tests assert the *rendered* Dockerfile: the check precedes the first copy, and its
path list matches what the renderer copies. What they cannot show is the message a human
actually sees, which is the thing being fixed. Not run here — no Docker daemon on this machine.

### 1. The local path, no Docker needed

```sh
uv run python -c "
from idegym.plugins.defaults.image import IdeGYMServer
from idegym.api.plugin import BuildContext
IdeGYMServer.from_local('/tmp').apply(BuildContext(base='debian:bookworm-slim'))
"
```

**Expect** a `ValueError` naming every missing path and telling you what the argument wants:

```
IdeGYM source at /tmp is missing: .python-version, api, backend-utils, common-utils,
entrypoint.py, ... IdeGYMServer.from_local(...) needs the root of an IdeGYM workspace.
```

**Failure looks like:** no exception at all — the check is not wired into `apply()`, and the
failure returns to happening mid-Docker-build.

### 2. The git path, with Docker

Build an image against a ref that predates `plugins/` (any commit from before that split):

```python
Image(base="debian:bookworm-slim").with_plugin(BaseSystem()).with_plugin(User(name="dev")).with_plugin(
    IdeGYMServer.from_git(url="https://github.com/JetBrains-Research/idegym.git", ref="<old-sha>")
).build()
```

**Expect** the build to stop at the check step, seconds after the clone, with:

```
IdeGYM source at https://github.com/JetBrains-Research/idegym.git@<old-sha> is missing: plugins
This ref predates the current workspace layout; pick a newer one.
```

**Failure looks like** the old behaviour: the build runs on past the check and dies later with
`cp: cannot stat '/tmp/idegym-src/plugins': No such file or directory` — naming neither the ref
nor the requirement, which is the afternoon this change is meant to save.

### 3. Readiness timeout wording (cluster)

Start a server with an image large enough not to be cached, and a short timeout:

```python
await client.start_server(image_tag="<multi-GB image>", server_start_wait_timeout_in_seconds=30)
```

**Expect** a `TimeoutError` ending in `still pulling the image or creating the container
(ContainerCreating)` — not a bare readiness timeout. **Failure looks like** the message stopping
at `(not waiting on the scheduler)`, which is what used to send readers at the health endpoint.

client/src/idegym/client/operations/servers.py
```diff
@@ -75,7 +75,7 @@ async def start_server(
         env_from: Optional[list[KubernetesEnvFromSource]] = None,
         service_account_name: Optional[str] = None,
         pod_overrides: Optional[KubernetesPodOverrides] = None,
-        server_start_wait_timeout_in_seconds: int = 60,
+        server_start_wait_timeout_in_seconds: int = 300,
```
samorojy: **Client deadline expires before the orchestrator's diagnostic can arrive.** The client wraps polling in `asyncio.timeout(remaining_time≈300)` (servers.py:138, and the same in `restart_server` at :229), and that clock starts earlier than the orchestrator's own `timeout(300)`, which only begins after the DB work and `deploy_server`, and is then followed by another `list_pods` call in `describe_pod_startup`. The result is that the SDK caller always gets the client's bare `TimeoutError` ("Server start timed out", or an empty asyncio timeout) before the operation fails with the new "(…; still pulling the image …)" message. The example error in `client.md` and `remote_deployment.md` never reaches user code. Consider giving the client some slack over the server-side timeout, e.g. timeout + a grace period.

samorojy: **The poll backoff leaves a large blind window near the new 300s deadline.** With the default `PollingConfig` (0.05s × 1.5ⁿ, capped at 120s), polls land at about 148s and 222s, and the next one is due at about 333s, after the 300s `asyncio.timeout`. If a cold pull finishes at around 230s, the orchestrator marks the server ALIVE and the operation SUCCEEDED, but the client stays asleep until 300s and then raises `TimeoutError`. The caller then retries or fails while a live server it owns keeps running until the inactivity cleanup. With the old 60s default this gap was only about 16s. Consider clamping each sleep to the remaining deadline, or doing one last poll right before giving up.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -508,10 +508,50 @@ def _budget_seconds(budget: Duration) -> float:
...
+        for container in (pod.status.container_statuses or [])
+        if container.state and container.state.waiting and container.state.waiting.reason
+    }
+    if waiting & _PULL_FAILED_REASONS:
+        return f"the image could not be pulled ({', '.join(sorted(waiting))})"
+    if waiting & _PULL_IN_PROGRESS_REASONS:
+        return f"still pulling the image or creating the container ({', '.join(sorted(waiting))})"
+    if waiting:
+        return f"phase {pod.status.phase}, container waiting ({', '.join(sorted(waiting))})"
+    if pod.status.phase == "Running":
```
samorojy: **"readiness probe has not passed" is reported without checking container readiness.** On a RESTART reuse, if the old pod is stuck terminating (long grace period or a finalizer) and the new pod is Running and ready, `pods_are_ready` returns `pods_ready=True, has_terminating_pods=True`, so the loop waits until it times out. Terminating pods are filtered out here, so this message blames the image's readiness probe, which is exactly the misattribution this PR is trying to remove. It would help to check `container_statuses[*].ready` and to name the terminating-pod case explicitly.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -508,10 +508,50 @@ def _budget_seconds(budget: Duration) -> float:
...
+    """
+    try:
+        pods = [pod for pod in await list_pods(label_selector, namespace) if pod.metadata.deletion_timestamp is None]
+    except Exception as error:  # noqa: BLE001  # a diagnostic must never replace the real failure
+        return f"pod state unavailable: {error}"
+
+    if not pods:
+        return "no pods matched"
+
+    pod = pods[0]
```
samorojy: **Only `pods[0]` is diagnosed.** With `nodes_count>1` (nodes_holder), or during a rollout where old and new non-terminating pods overlap, pod 0 can be Running and ready while another pod is still in ContainerCreating on a cold node. The message then says "container running, but its readiness probe has not passed" when it should say that a pod is still pulling. Prefer the first pod that is *not* ready, or summarise across all pods.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -508,10 +508,50 @@ def _budget_seconds(budget: Duration) -> float:
     return budget.total_seconds() or inf
 
 
+# What the kubelet reports on a container that has not started yet. A pull in progress shows up
+# as ContainerCreating — there is no distinct "Pulling" waiting reason — which is why a slow pull
+# and a broken readiness probe are indistinguishable without looking at this.
+_PULL_IN_PROGRESS_REASONS = frozenset({"ContainerCreating", "PodInitializing"})
```
samorojy: **`PodInitializing` is treated as a pull in progress, and `init_container_statuses` are ignored.** If `pod_overrides` adds an init container that is crash-looping or blocked on a volume, the main container sits in `waiting=PodInitializing`, and the timeout reads "still pulling the image or creating the container (PodInitializing)". The user is told to raise the timeout, and retrying hits the same failure. Consider looking at `init_container_statuses` first and reporting a waiting or terminated init container by name.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -588,9 +628,12 @@ async def wait_for_pods_ready(
     except TimeoutError:
         # The budgets are routinely larger than a caller's overall timeout, so this — not the
         # RuntimeError above — is how a pod that never schedules usually surfaces. Say what the
-        # scheduler was doing, or the failure is an unattributable timeout.
+        # scheduler was doing and what the pod itself was doing, or the failure is an
+        # unattributable timeout that reads as a broken health endpoint.
+        startup_state = await describe_pod_startup(label_selector, namespace)
```
samorojy: **The diagnostic makes another K8s API call after the timeout budget has run out, with no time limit.** When the API server is slow or unreachable, which is often why the wait timed out in the first place, `describe_pod_startup` opens a fresh `async_kube_api()` and awaits `list_namespaced_pod` with no timeout. The `TimeoutError` then arrives well after `wait_timeout`, or the call hangs. Consider reusing the pod list from the last `pods_are_ready` poll, or wrapping this call in a short `asyncio.timeout`.

unit-tests/test_idegym_server_source_check.py
```diff
@@ -0,0 +1,98 @@
...
+    assert check_at < first_copy_at
+
+
+def test_the_git_check_names_the_url_and_the_ref() -> None:
+    dockerfile = IdeGYMServer.from_git(url="https://example.test/idegym.git", ref="abc123").render(_context())
+
+    assert "IdeGYM source at https://example.test/idegym.git@abc123 is missing" in dockerfile
+
+
+def test_the_git_check_covers_every_path_the_renderer_copies() -> None:
```
samorojy: **This test can't detect the drift its name says it guards against.** It only compares the rendered loop to `_REQUIRED_WORKSPACE_PATHS` itself. If someone adds `cp -r /tmp/idegym-src/newpkg` and `COPY newpkg newpkg/` without updating the tuple, every test still passes, and a stale ref goes back to failing with a bare `cp: no such file`. Either assert against the actual `cp` and `COPY` lines in the rendered Dockerfile, or, more simply, generate those lines from `_REQUIRED_WORKSPACE_PATHS` so there is only one list.

plugins/defaults/src/idegym/plugins/defaults/image.py
```diff
@@ -535,6 +535,52 @@ def render(self, ctx: BuildContext) -> str:
...
+    checks = " ".join(quote(path) for path in _REQUIRED_WORKSPACE_PATHS)
+    return dedent(
+        f"""\
+        RUN set -eu; \\
+            missing=""; \\
+            for path in {checks}; do \\
+                [ -e {quote(source_root)}/"$path" ] || missing="$missing $path"; \\
+            done; \\
+            if [ -n "$missing" ]; then \\
+                echo "IdeGYM source at {described_as} is missing:$missing" >&2; \\
```
samorojy: **url and ref are interpolated unquoted into a double-quoted `echo`.** The clone line uses `quote(self.url)` / `quote(ref)`, but here a url such as `https://u:pa$$w0rd@host/idegym.git`, or a ref containing `"` or `$(...)`, either breaks the RUN step with a shell syntax error or gets expanded or executed inside the error message. Consider building `described_as` from the `shlex.quote`d values, or passing it through `printf '%s'`.

backend-utils/src/idegym/backend/utils/kubernetes_client.py
```diff
@@ -508,10 +508,50 @@ def _budget_seconds(budget: Duration) -> float:
     return budget.total_seconds() or inf
 
 
+# What the kubelet reports on a container that has not started yet. A pull in progress shows up
+# as ContainerCreating — there is no distinct "Pulling" waiting reason — which is why a slow pull
+# and a broken readiness probe are indistinguishable without looking at this.
+_PULL_IN_PROGRESS_REASONS = frozenset({"ContainerCreating", "PodInitializing"})
+_PULL_FAILED_REASONS = frozenset({"ImagePullBackOff", "ErrImagePull", "InvalidImageName"})
```
samorojy: nit: `_PULL_FAILED_REASONS` duplicates the inline `["ImagePullBackOff", "ErrImagePull"]` list in `pods_are_ready`, and the two already differ. `InvalidImageName` is reported as "could not be pulled", but it never triggers the fail-fast `RuntimeError`, so the wait burns the full 300s on an image name that will never resolve. `pods_are_ready` should use this frozenset. The logic that filters out terminating pods and then takes `pods[0]` also repeats `pod_phase_and_readiness`.

website/docs/reference/image_builder.md
```diff
@@ -7,8 +7,18 @@ IDEs — and the builder assembles the Dockerfile from reusable, validated build
 Images can be defined in Python (fluent API) or YAML, and can be built locally with Docker or inside the
 cluster with Kaniko.
 
+## Installation
```
samorojy: nit: This "Installation" section looks unrelated to image source validation and start timeouts. AGENTS.md asks to keep the diff in scope, since the squash commit becomes a changelog entry. Could it go in a separate PR?

samorojy (approved):
Approving. The direction is good, and pod-state diagnosis on start timeouts is a real improvement. I've left some inline comments. The main one: the client's own deadline (plus the poll backoff) currently fires before the orchestrator's diagnostic can reach SDK callers, so the new message isn't visible yet in practice. The rest are about the diagnosis being more precise (multi-pod, init containers, terminating pods, bounding the extra API call) and a quoting fix in the source-check `echo`.

PR #297, opened 2026-09-01
[ci][deps] Run checks on any PR base; migrate to fastmcp 4

vpoliakov-pixel:
Two commits, both prerequisites for the JBRes-10676 stack (#289 ← #290 ← #291 ← #292 ← #293 ← #294 ← #295 ← #296). Merge this first; without it seven of those eight PRs have **zero** checks and stay blocked.

## What changed

`ci.yml` and `lint.yml` were filtered to `pull_request: branches: [main]`, so a pull request based on another feature branch got no checks at all — not lint, not unit, not integration. Dropping the filter fixes that. `push` stays `main`-only, so pushing a feature branch does not double-trigger.

**The six e2e groups stay gated** on a `main` base (`if: github.event_name != 'pull_request' || github.base_ref == 'main'`). Each provisions its own cluster, so running all six on every layer of a stack costs far more than it tells us. A layer gets lint + unit + integration — which includes the database round-trip, the part most worth catching early — and picks up e2e automatically when its base merges, because **GitHub retargets a pull request onto its base's base** when the base is merged.

Branch protection is unaffected: it applies to `main`, and a PR targeting `main` still runs everything. Intermediate merges in a stack target unprotected feature branches. No check is renamed, so no required-check configuration changes.

`docs.yml` had a single global `pages` concurrency group. Eight stacked PRs queued in it and GitHub kept only the newest pending run, cancelling the rest — which is why that stack produced seven cancelled docs builds. Validation builds now take a lane per branch; the production deploy keeps the global lane it needs, and the deploy job is already `main`-push-only.

## How to verify

This PR is its own test: it targets `main`, so it must show the full check set including all six e2e groups. Then on any stacked PR — e.g. #290, currently showing "no checks reported" — the checks should appear as `Lint / Python`, `CI / Run unit tests`, `CI / Run integration tests`, `Build site`, with the e2e groups absent.

**Failure looks like:** e2e groups appearing on a non-`main`-base PR (the `if` is wrong), or missing on this one (the `if` is inverted). All three workflow files parse as YAML.

## Second commit: migrate to fastmcp 4

`e2e openhands` has been red on every PR in the repo since 2026-08-31, including this branch when it contained no Python at all. Cause: FastMCP 4.0.0 was published that day, and the OpenHands service venv is the **only** install here that bypasses `uv.lock` — the image plugin ran `uv pip install … "fastmcp>=3"` — so it adopted the new major on its own. FastMCP 4 moved `fastmcp.tools.tool` to `fastmcp.tools`, so `service/mcp.py` failed at import, nothing bound the loopback port, and every call answered `503 service_unavailable`.

Migrated rather than pinned back: import updated, `server`/`orchestrator` require `fastmcp>=4.0.0`, `uv.lock` resolves 4.0.0. The diff is small because everything else survives the major unchanged — `FastMCP`, `Client`, `ToolError`, `combine_lifespans`, `server.create_proxy`, `http_app(path=…)`, `add_tool`, and subclassing `Tool` with an overridden `run(self, arguments)` all verified identical on 4.0.0.

Two consequences to review:

- FastMCP 4 needs `mcp` 2.x, which moves the MCP stack onto **`httpx2`** — a separate distribution from `httpx`, not an upgrade. Both are in the lock and import side by side; IdeGYM's own clients stay on `httpx 0.28.1`. Documented in `COMPATIBILITY.md` so nobody tries to consolidate them.
- MCP SDK v2 renamed `Tool.inputSchema` → `input_schema`; the two assertions reading it are updated. The `inputSchema` **dict keys** in the OpenHands adapters are the wire format and deliberately untouched.

**Verified locally:** `ruff check` clean, 390 files formatted, `uv lock --check` clean, **1162 unit passed** with zero FastMCP deprecation warnings, 46 docker-free integration passed. The real proof is `Run e2e openhands tests` on this PR.

## Original control-experiment note

The e2e run on this PR is a **control experiment**. `CI / Run e2e openhands tests` is failing on #289 and reproduced on a re-run of the unchanged commit; #289 touches only the bash executor, so I want to know whether that job is broken independently of my changes. This branch contains no Python at all — if openhands fails here too, it is pre-existing.

Resolves: nothing on its own — unblocks JBRes-10676.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,9 +3,9 @@ on:
   push:
     branches:
       - main
+  # No branch filter: a stacked pull request based on another feature branch needs checks too,
+  # or every layer above the first is unverifiable. The e2e job below is what stays gated.
   pull_request:
-    branches:
-      - main
 
 jobs:
   test:
@@ -67,6 +67,13 @@ jobs:
           IDEGYM_TEST_REGISTRY: ${{ matrix.registry }}
 
   e2e:
+    # Reserved for changes that are actually landing: a pull request targeting main, or a push to
+    # main. Each matrix group provisions its own cluster, so running six of them on every layer of
+    # a stack costs far more than it tells us — a stacked pull request gets lint, unit and
+    # integration (which covers the migration round-trip), and the full suite runs when the stack
+    # reaches main. GitHub retargets a pull request onto its base's base when the base merges, so
+    # each layer picks up e2e automatically as the stack lands.
+    if: github.event_name != 'pull_request' || github.base_ref == 'main'
     # One matrix entry per balanced e2e group. Every group recreates the same cluster once and
     # differs only by the pytest ``-m`` marker expression. The "other" group is the catch-all so a
     # new e2e test file is never silently dropped. Group markers are auto-applied by ``conftest.py``.
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -23,9 +23,12 @@ permissions:
   pages: write
   id-token: write
 
-# Allow only one concurrent deployment; don't cancel an in-progress prod deploy.
+# `pages` is a single global lane so a production deploy is never raced. A validation build on a
+# pull request needs its own lane per branch: several stacked pull requests otherwise queue in the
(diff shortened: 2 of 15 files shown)


DaniilMironovJB (approved):
🚀

PR #298, opened 2026-09-02
[JBRes-10600] Build from an inline Dockerfile base with a GCS build context

DaniilMironovJB:
An `Image` can now name its base as Dockerfile text (`base_dockerfile`) instead of a registry
reference, compiled into the same build as the plugin stages. That removes the separate
build-and-push cycle a custom base used to need, along with its registry write credentials, the
second async build the orchestrator never tracked, and the intermediate image nobody owns.

`context_uri` points the build at an archive the caller has already staged, so an inline base's
`COPY`/`ADD` resolve without the orchestrator ever receiving context bytes. Also adds
Secret Manager-backed `secrets`, a caller-supplied destination tag gated on a registry allowlist,
and per-request build resources.

The merge emits the author's text verbatim apart from one edit: the stage acting as the base gains
an `AS` alias, and only if it does not already declare one — renaming it would break their own
`COPY --from=` references. `FROM <alias>` then inherits that stage's image config, which is what
makes this equivalent to publishing the base and referencing it by tag; a test asserts the
generated segment is byte-identical between the two forms.

## How it works

`_render_dockerfile()` concatenates the sections below in this order, dropping the empty ones. The
first two are what an inline base adds; everything from `FROM` down is identical whichever base
form was used, which is what makes the two equivalent.

```mermaid
flowchart LR
  subgraph def["Image definition"]
    bd["base_dockerfile (new)"]
    b["base"]
    pl["plugins"]
    cmd["commands"]
  end

  subgraph out["Rendered Dockerfile, top to bottom"]
    d1["parser directives, hoisted (new)"]
    d2["author stages, verbatim plus AS idegym_base (new)"]
    d3["plugin build stages"]
    d4["FROM base or alias, SHELL, USER root"]
    d5["IDEGYM_ project and auth ARG and ENV block"]
    d6["plugin fragments"]
    d7["USER ctx.current_user"]
    d8["RUN, the chained commands"]
  end

  bd --> d1
  bd --> d2
  bd -. "stage alias" .-> d4
  b -. "registry reference" .-> d4
  pl --> d3
  pl --> d6
  pl -. "ctx.request, project_root" .-> d5
  pl -. "ctx.current_user" .-> d7
  cmd --> d8
```

The author's text is emitted verbatim apart from the `AS` alias, and the plugin fragments land
*after* `FROM`, which is why a base's `ENTRYPOINT`/`CMD`/`HEALTHCHECK` is the thing that gets
overridden.

### Components and where each check lives

One pure-text analysis module underpins the whole path, and every gate runs before a build is
submitted so a bad definition fails on the request rather than in a build log minutes later.

```mermaid
flowchart TB
  an["dockerfile_analysis (new): logical lines, stages, COPY sources, BuildKit-only features"]
  def["Image validators: exactly one base, known base_stage, no reserved idegym_ stage, no IDEGYM_AUTH_TOKEN"]
  spec["to_spec(): normalize and alias the base stage, require a context for COPY, record override warnings"]
  model["ImageBuildSpec: shapes of context_uri, tag, registry, version, secret mapping"]
  svc["ImageBuildService.resolve_tag: registry allowlist (new)"]
  kan["kaniko submit: reject BuildKit-only syntax, unfetchable scheme, two contexts; secrets become build args (new)"]
  cbg["cloudbuild_gke submit: gs:// only, clamp timeout and disk, allowlist machine type, inject BUILDKIT_SYNTAX, overlay the context (new)"]
  loc["local docker driver: reject context_uri, secrets and tag outright (new)"]
  job["job record details: spec plus handle warnings (new)"]

  an --> def
  an --> spec
  def --> spec
  spec --> model
  model --> svc
  model --> loc
  svc --> kan
  svc --> cbg
  an --> kan
  an --> cbg
  kan --> job
  cbg --> job
```

### Three deviations from the ticket, each deliberate

- **Cloud Build overlays the caller context** with a fetch step rather than using `context_uri` as
  the `StorageSource` (Part 3). The StorageSource carries the generated Dockerfile and the plugin
  assets, so swapping it drops both.
- **`context_uri` + plugin `context_files` is rejected on Kaniko.** Kaniko accepts one `--context`
  and the plugin assets already use it. Works on Cloud Build; validated up front and documented.
- **`BUILDKIT_SYNTAX` is injected only when the Dockerfile needs it**, not unconditionally
  (Part 5). Unconditional injection makes every Cloud Build pull `docker/dockerfile:1` from Docker
  Hub, adding a rate limit and an egress dependency to builds that never needed one.

### Fixes two pre-existing bugs on `cloudbuild_gke`

Both live in the functions this change touches, and both would have left the first consumer broken:
`spec.context_files` were dropped from the context tar, so any image using the `idea`/`pycharm`
plugins could not build there at all; and an Artifact Registry tag was looked up as a digest, so
`skip_existing` never skipped anything.

### Worth attention

- `ENTRYPOINT`, `CMD` and `HEALTHCHECK` declared by the base are overridden by `idegym-server`.
  Not new — a pre-published `base:` behaves identically — but far more likely to bite here. Now
  warned about on the job record and documented.
- `Image.project_root` still defaults to `/root/work` and is only settable through the `Project`
  plugin, which also clones. A base that clones into its own `WORKDIR` therefore leaves the agent
  pointed elsewhere. Not addressed here; needs its own decision.
- `build_args` (Part 11) is deliberately **not** implemented — the ticket recommended deferring it
  until a concrete need appears, and none of the current consumers have one.

Verified against real Docker, a real Kaniko build on minikube, and six real Cloud Builds; the
`.dockerignore` behaviour is pinned by a test that fails without the fix.

Resolves: JBRes-10600

## Checklist


vpoliakov-pixel:
Would be great to have a diagram of some sort describing the changes

api/src/idegym/api/dockerfile_analysis.py
```diff
@@ -0,0 +1,321 @@
+"""Pure text analysis of Dockerfile content.
+
+Lives in ``api`` because both build backends need it and ``backend-utils`` depends on ``api``
```
vpoliakov-pixel: We have common-utils for that, no need to put actual code in api module

api/src/idegym/api/dockerfile_analysis.py
```diff
@@ -0,0 +1,321 @@
...
+alone; `idegym.image.base_dockerfile` builds its normalization on the same primitives. No
+Pydantic, no I/O.
+
+Scanning happens at *logical* line level, the way Docker parses: continuations are joined and
```
vpoliakov-pixel: To my taste, the whole docstring for this file could be removed

api/src/idegym/api/image_build.py
```diff
@@ -17,6 +18,176 @@ class BuildBackend(StrEnum):
...
+        raise ValueError(f"Build context URI {value!r} is not well formed")
+    return match.group("scheme").lower()
+
+
+# A Dockerfile ``ARG`` name. Secret ids are held to the same pattern: Kaniko passes each one as a
+# build arg, and a name containing spaces or ``=`` could inject extra arguments there.
+_BUILD_ARG_NAME_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+# ``projects/<project>/secrets/<secret>``, optionally pinned to ``/versions/<version>``.
+_SECRET_RESOURCE_RE = re.compile(r"^projects/[^/\s]+/secrets/[^/\s]+(?:/versions/[^/\s]+)?$")
```
vpoliakov-pixel, line 53: Either I do not understand what it means or it won't "every" possible location of secrets..

backend-utils/src/idegym/backend/utils/image_builder/cloudbuild_gke.py
```diff
@@ -69,6 +90,21 @@ def build_cloudbuild_config(
         if spec.request.auth.token is not None:
             docker_args += ["--secret", f"id={AUTH_SECRET_ID},src=./{AUTH_SECRET_SRC}"]
 
+    # Plugin-declared secrets, resolved from the orchestrator's own environment as the Kaniko
+    # backend has always done; this one used to drop them, silently breaking `external_plugins`.
```
vpoliakov-pixel: Sounds not cool :)

backend-utils/src/idegym/backend/utils/image_builder/cloudbuild_gke.py
```diff
@@ -80,35 +116,137 @@ def build_cloudbuild_config(
...
+    input.
+
+    ``exclude`` is appended to ``.dockerignore`` after extraction, the one exception to
+    generated-files-win: shipping ours would discard the caller's, sweeping files they deliberately
+    excluded into the image via a broad ``COPY .``.
+    """
+    archive = "/tmp/idegym-build-context.archive"
+    commands = [
+        "set -eu",
+        f"gcloud storage cp {quote(context_uri)} {archive}",
```
vpoliakov-pixel, line 183: We are trying to avoid using explicit cli tools; try to perform the same using python gcloud client?

backend-utils/src/idegym/backend/utils/image_builder/kaniko.py
```diff
@@ -42,6 +48,40 @@ def _kaniko_git_context(version: str) -> str:
     return f"git://{url}#{_kaniko_git_ref(version)}"
 
 
+def validate_kaniko_spec(spec: ImageBuildSpec) -> None:
+    """Reject a spec Kaniko cannot build, before any Job is created.
+
+    Otherwise each surfaces minutes later as a build-log failure that reads as an infrastructure
```
vpoliakov-pixel: Do not really need this comment

api/src/idegym/api/image_build.py
DaniilMironovJB: Discussed offline, this is google secret manager paths

vpoliakov-pixel (approved):

PR #303, opened 2026-09-11
[JBRes-10125] Preserve task JDK and recover build status

DaniilMironovJB:
Preserve the base image's `JAVA_HOME` and `PATH` precedence when installing IDEA or PyCharm. The IDE continues to use its bundled runtime.

Persist each build's **build context** -- `<backend>://<location>`, where the scheme is a `BuildBackend` value and the location is that backend's own addressing -- so the watcher can recover Cloud Build and Kaniko status after an orchestrator restart. This addresses a build where it was submitted rather than where the deployment currently builds, so switching backends does not strand in-flight builds. Includes database migration `004` and regression tests for Java environment preservation, status recovery, and migration rollback.

Deployment: the watcher needs `cloudbuild.builds.get` in the build project.

**Upgrade note:** the watcher reconciles a build only through the backend its persisted build context names. Builds already `IN_PROGRESS` when this release rolls out have no context and are left to the orchestrator's own monitor; one orphaned by a restart in that window stays `IN_PROGRESS` until it ages out.

Related: JetBrains-Research/agent-eval-runner#59
Ticket: JBRes-10125

## Checklist

- [x] `uv run ruff check` and `uv run ruff format --check` both pass.
- [x] Tests cover the change at the unit and database integration levels.
- [x] `AGENTS.md` updated with the Java environment preservation invariant.


backend-utils/src/idegym/backend/utils/image_builder/status.py
```diff
@@ -0,0 +1,19 @@
+from idegym.api.status import Status
+from idegym.backend.utils.kubernetes_client import get_job_status
+
+
+async def get_build_status(resource: str) -> Status:
+    """Query a persisted build resource without relying on deployment defaults."""
+    backend, location = resource.split("://", 1)
+    if backend == "kaniko":
```
vpoliakov-pixel: We have BuildBackend enum and it has a bit different values..

orchestrator/src/idegym/orchestrator/image_build_service.py
```diff
@@ -113,6 +112,7 @@ async def monitor_image_building_job(
                     status=Status.IN_PROGRESS,
                     tag=tag,
                     details=details,
+                    build_resource=handle.resource,
```
vpoliakov-pixel: Personally I would prefer using "context" instead of "resource" as it is used in cloudbuild for consistency

website/docs/reference/image_builder.md
```diff
@@ -1236,3 +1236,11 @@ pass data between plugins via `extras`.
 > **See also:** [Plugin Architecture](plugins.md) — full guide covering server plugins, client
 > operation plugins, MCP upstream convention, the `plugins.json` configuration file, and how to
 > write a plugin that participates in all integration points.
+
+
+### Build reconciliation
+
+The watcher uses each build's persisted backend resource to recover completion after an
+orchestrator restart. Its identity needs `cloudbuild.builds.get` for the build project.
+Older records without a resource are reconciled only for the historical `kaniko-build-`
```
vpoliakov-pixel: Do we really need to separate it? Legacy and new builds? Lets not increase the amount of corner cases and just mention it explicitly in release changelog

vpoliakov-pixel (approved):

